PR #16, opened 2019-06-26
Update PL-SF to work with PennyLane v0.4 and SF v0.10

josh146:
**Description of the Change:** this PR updates the plugin to work with v0.4 of PennyLane, and the current release v0.10 of Strawberry Fields.

**Benefits:** Compatibility

**Possible Drawbacks:** n/a

### DIFF
--- a/pennylane_sf/simulator.py
+++ b/pennylane_sf/simulator.py
@@ -22,8 +22,8 @@ Base simulator class
 A base class for constructing Strawberry Fields devices for PennyLane.
 This class provides all the boilerplate for supporting PennyLane;
 inheriting devices simply need to provide their engine run command
-in :meth:`~.StrawberryFieldsSimulator.pre_expval`, as well as defining their ``_operation_map``
-and ``_expectation_map``, mapping PennyLane operations to their
+in :meth:`~.StrawberryFieldsSimulator.pre_measure`, as well as defining their ``_operation_map``
+and ``_observable_map``, mapping PennyLane operations to their
 Strawberry Fields counterparts.
 
 Classes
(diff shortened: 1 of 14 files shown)


pennylane_sf/simulator.py
```diff
@@ -122,6 +123,20 @@ def expval(self, expectation, wires, par):
...
+
+        Args:
+            observable (str): name of the observable
+            wires (Sequence[int]): subsystems the observable is evaluated on
+            par (tuple): parameters for the observable
+
+        Returns:
+            float: variance value
+        """
+        _, var = self._observable_map[observable](self.state, wires, par)
```
co9olguy: just curious, is it more efficient to set up separate functions for expectation values and variances, rather than a single function which returns both, even if only one is needed?

tests/test_fock.py
```diff
@@ -355,11 +355,54 @@ def SF_reference_trace(x, y):
...
+            qml.Squeezing(r, 0, wires=0)
+            qml.Rotation(phi, wires=0)
+            return qml.var(qml.X(0))
+
+        r = 0.543
+        phi = -0.654
+
+        var = circuit(r, phi)
+        expected = np.exp(2*r)*np.sin(phi)**2 + np.exp(-2*r)*np.cos(phi)**2
+        assert np.allclose(var, expected, atol=1e-3, rtol=0)
```
co9olguy: magic numbers?

pennylane_sf/simulator.py
josh146: This is an SF limitation --- all state methods return a tuple of expectations and variance, so it doesn't sync well with Pennylane's one-or-the-other approach. Worth making an issue for this and seeing if there is a better solution.

tests/test_fock.py
josh146: Hmm yeah I don't like this either. Just an issue with cutoff dimension - I could increase it beyond 15 to get better accuracy, but then the test takes longer. I'll try decreasing the mean photon number.

co9olguy: yup that's easiest

pennylane_sf/simulator.py
co9olguy: k, so the efficiency gain would be found within SF itself, not plugin. Got it.

tests/test_fock.py
josh146: done

PR #17, opened 2019-07-06
Update PennyLane-SF for breaking changes in SF v0.11

josh146:
**Description of the Change:** Updates the PennyLane-SF plugin for the breaking changes introduced in SF version 0.11

**Benefits:** PennyLane-SF now works with latest development version of SF.

**Possible Drawbacks:** n/a


josh146:
This has uncovered a bug in SF on Python 3.5, as `ModuleNotFoundError` is not defined. However, the tests pass on 3.6.

josh146:
@co9olguy, this is now ready for review. I've temporarily moved travis from testing against python 3.5 and 3.6 to 3.6 and 3.7 to avoid this bug, as it will instead need to be fixed in SF.

requirements.txt
```diff
@@ -1,2 +1,2 @@
-strawberryfields>=0.9
-pennylane>=0.4
+strawberryfields
```
co9olguy: no versions?

setup.py
```diff
@@ -22,8 +22,8 @@
 
 
 requirements = [
-    "strawberryfields>=0.9",
-    "pennylane>=0.4"
+    "strawberryfields",
```
co9olguy: same question

requirements.txt
josh146: You're right, this is an oversight on my end.
```suggestion
strawberryfields>=0.11
```

josh146: Ah, I remember now. Travis is set to install SF from github master - setting >=0.11 before 0.11 was released was causing pip to fail, so I temporarily removed version pinning.

setup.py
josh146:
```suggestion
    "strawberryfields>=0.11",
```

requirements.txt
```diff
@@ -1,2 +1,2 @@
-strawberryfields>=0.9
-pennylane>=0.4
+strawberryfields>=0.11
+pennylane
```
josh146:
```suggestion
pennylane>=0.4
```

setup.py
```diff
@@ -22,8 +22,8 @@
 
 
 requirements = [
-    "strawberryfields>=0.9",
-    "pennylane>=0.4"
+    "strawberryfields>=0.11",
+    "pennylane"
```
josh146:
```suggestion
    "pennylane>=0.4"
```

PR #22, opened 2019-09-09
Updates the plugin for the 0.5 release

josh146:

* Fixes tests to ensure they are compatible with the new `wires.Any` and `wires.All` enumerations.

* Updates the version number

**Benefits:** n/a

**Possible Drawbacks:** n/a

### DIFF
--- a/tests/test_gaussian.py
+++ b/tests/test_gaussian.py
@@ -65,7 +65,7 @@ class GaussianTests(BaseTest):
         for g in all_gates - gates:
             op = getattr(qml.ops, g)
 
-            if op.num_wires == 0:
+            if op.num_wires <= 0:
                 wires = [0]
             else:
                 wires = list(range(op.num_wires))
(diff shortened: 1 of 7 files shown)


tests/test_gaussian.py
```diff
@@ -65,7 +65,7 @@ def test_unsupported_gates(self):
         for g in all_gates - gates:
             op = getattr(qml.ops, g)
 
-            if op.num_wires == 0:
+            if op.num_wires <= 0:
```
co9olguy: just curious, why are these changed?

josh146: Originally PennyLane operations either specified a specific number of required wires, or 'all wires' (`Operation.num_wires = 0`).

In practice though, this didn't work, as there existed operations that didn't fit into these two categories --- namely operations that work on 'any' wires (`QubitStateVector`, `BasisState`, `GaussianState`).

In https://github.com/XanaduAI/pennylane/pull/277, we created enumerations for `Any` and `All` wires to fix this. For backwards compatibility, `All == 0`, and `Any = -1`.

This test is marked for deprecation, so this is just a quick fix to ensure it continue to pass.

co9olguy: That's what I figured. :+1:

Issue #23, opened 2019-09-26
Device 'Shots=0' argument breaks down for the SF plugin


co9olguy:
The SF plugin has not been updated with PL Device reconfiguration where the user is not explicitly required to give 'shots=0' argument anymore. This produces the following bug when running files that use SF plugin:

```pennylane._device.DeviceError: The specified number of shots needs to be at least 1. Got 0.```

*See https://github.com/XanaduAI/pennylane/issues/334*

**Comment by AroosaIjaz** [comment]
@josh146, even after reinstalling the plugin, if I run `make docs` on the master PL branch, I still see the same error:

```
Unexpected failing examples:
/home/aroosa/pennylane/examples/pennylane_run_plugins_hybrid.py failed leaving traceback:
Traceback (most recent call last):
  File "/home/aroosa/conda3/lib/python3.7/site-packages/sphinx_gallery/gen_rst.py", line 480, in _memory_usage
    out = func()
  File "/home/aroosa/conda3/lib/python3.7/site-packages/sphinx_gallery/gen_rst.py", line 465, in __call__
    exec(self.code, self.globals)
  File "/home/aroosa/pennylane/examples/pennylane_run_plugins_hybrid.py", line 131, in <module>
    dev_fock = qml.device("strawberryfields.fock", wires=2, cutoff_dim=2)
  File "/home/aroosa/pennylane/pennylane/__init__.py", line 204, in device
    return plugin_device_class(*args, **options)
  File "/home/aroosa/conda3/lib/python3.7/site-packages/pennylane_sf/fock.py", line 100, in __init__
    super().__init__(wires, shots=shots, hbar=hbar)
  File "/home/aroosa/conda3/lib/python3.7/site-packages/pennylane_sf/simulator.py", line 68, in __init__
    super().__init__(wires, shots)
  File "/home/aroosa/pennylane/pennylane/_device.py", line 121, in __init__
    self.shots = shots
  File "/home/aroosa/pennylane/pennylane/_device.py", line 194, in shots
    raise DeviceError("The specified number of shots needs to be at least 1. Got {}.".format(shots))
pennylane._device.DeviceError: The specified number of shots needs to be at least 1. Got 0.
```
It might be a plugin update issue. Should I update from the github repo of the plugin? I just redid the `pip install ` command.

co9olguy:
Yes, the update has not propagated into a release of pennylane-sf

co9olguy:
[...]

PR #24, opened 2019-09-26
Adds analytic argument, removes shots=0

josh146:

* Adds `analytic=True` as a default argument
* Changes `shots=0` to `shots=1000` by default

**Benefits:**

Brings the plugin inline with https://github.com/XanaduAI/pennylane/pull/317

**Possible Drawbacks:**

None. Note that the SF plugin currently does not support sampling, so the ``shots`` argument only affects expectation/variance estimation.

**Related GitHub Issues:** #23

### DIFF
--- a/tests/test_fock.py
+++ b/tests/test_fock.py
@@ -52,7 +52,7 @@ class FockTests(BaseTest):
         self.assertEqual(dev.num_wires, 2)
         self.assertEqual(dev.cutoff, 5)
         self.assertEqual(dev.hbar, 2)
-        self.assertEqual(dev.shots, 0)
+        self.assertEqual(dev.shots, 1000)
         self.assertEqual(dev.short_name, 'strawberryfields.fock')
 
     def test_fock_args(self):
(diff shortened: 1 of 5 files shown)


tests/test_fock.py
```diff
@@ -75,14 +75,15 @@ def test_unsupported_gates(self):
         for g in all_gates - gates:
             op = getattr(qml.ops, g)
 
-            if op.num_wires == 0:
+            if op.num_wires <= 0:
```
co9olguy: We might as well use the enums here. We made them for this purpose 😆

tests/test_fock.py
```diff
@@ -75,14 +75,15 @@ def test_unsupported_gates(self):
...
                 wires = [0]
             else:
                 wires = list(range(op.num_wires))
 
             @qml.qnode(dev)
             def circuit(*args):
                 args = prep_par(args, op)
+                print(wires)
```
co9olguy: remove

tests/test_fock.py
```diff
@@ -106,7 +107,7 @@ def test_unsupported_expectations(self):
         for g in all_obs - obs:
             op = getattr(qml.ops, g)
 
-            if op.num_wires == 0:
+            if op.num_wires <= 0:
```
co9olguy: see above about enums

tests/test_fock.py
josh146: Oops, good catch.

tests/test_fock.py
josh146: Done!

PR #37, opened 2020-06-10
Number expectation

antalszava:
Adds the ability to compute the expectation value of tensor number operators. For this, the `number_expectation` method of state objects is being used from Strawberry Fields.

Allowing the calculation of the variance of tensor number operators and
specifying more than 2 modes for the Gaussian device would be potential
additions.

**Benefits:**
- The expectation value of the `TensorN` observable can be used in circuits

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A


pennylane_sf/expectations.py
```diff
@@ -94,6 +94,23 @@ def mean_photon(state, wires, params):
...
+        state (strawberryfields.backends.states.BaseState): the quantum state
+        wires (Sequence[int]): the sequence of modes to measure
+        params (Sequence): sequence of parameters (not used)
+
+    Returns:
+        float, float: the expectation value of the number operator and its
+            variance
+    """
+    # pylint: disable=unused-argument
+    return state.number_expectation(wires)
```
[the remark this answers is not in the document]
antalszava: Just realized when trying to go for this, that we inherently return the `NumberOperator` for `TensorN` when constructed on a single mode. :smile:

antalszava:
Thanks for the comments @josh146, they were really helpful! :blush: if all's true implemented the suggestions. For a single mode `TensorN`, got the surprise that we'd inherently get a `NumberOperator`, and resort to `mean_photon` :)

pennylane_sf/expectations.py
```diff
@@ -94,6 +94,23 @@ def mean_photon(state, wires, params):
...
+        state (strawberryfields.backends.states.BaseState): the quantum state
+        wires (Sequence[int]): the sequence of modes to measure
+        params (Sequence): sequence of parameters (not used)
+
+    Returns:
+        float, float: the expectation value of the number operator and its
+            variance
+    """
+    # pylint: disable=unused-argument
+    return state.number_expectation(wires), None
```
josh146, line 111: @antalszava, did you want to add an if statement here?
```python
if len(wires) == 1:
    return mean_photon(state, wires, params)
return state.number_expectation(wires), None
```

The only difference will be that if it's only acting on 1 mode, variance will be supported. But this is minor.

antalszava: Right, yes, this is a great idea! The reason why I haven't added it, is because by this step of the execution it has already been taken care of. As [`TensorN` for `len(wires) == 1` returns the  `NumberOperator` ](https://github.com/XanaduAI/pennylane/blob/bcc4de1b2af476ed0e05ea0ded54ab57a3cd3754/pennylane/ops/cv.py#L860), upon instantiation of `TensorN` we get the `NumberOperator` and then here in PennyLane-SF we already deal with a `NumberOperator` instance. Therefore this function is called when `len(wires) != 1`.

josh146: Oh, good point! Disregard my comment in that case, I had forgotten we had made that decision in PennyLane core :)

PR #39, opened 2020-06-19
Added support for qml.probs to the simulator devices.

josh146:

* Adds a new `StrawberryFieldsSimulator.probability()` method that returns the probability (up to a defined cutoff) of the state or reduced state.

* Fixes the `expectations.py` file, by changing `GaussianState` to `BaseGaussianState`, and removing deprecated code.

* Changes `requirements.txt` to install master Strawberry Fields

**Benefits:**

* Enables support for `qml.prob` for both the Fock and Gaussian devices

* Finite difference _should_ work, but need to check it. Parameter shift shouldn't work! Also need to check this.

  Edit: looks like we need to add the following to the `BaseQNode` in PennyLane-Core:

  ```python
  elif res.return_type is ObservableReturnTypes.Probability:
      self.output_conversion = np.squeeze
      self.output_dim = 2 ** len(res.wires)

      if self.model == "cv":
          self.output_dim = getattr(self.device, "cutoff", 10)
  ```

**Possible Drawbacks:**

* The `pennylane.Device` class still requires that the probability be returned as an ordered dictionary. (EDIT Antal: added this to our list of potential changes to the Device class)

* The `gaussian` device should now accept `cutoff` as an _optional_ keyword argument on device load.

* It looks like there is a bug in SF master --- SF won't import if TF is not installed.

**Related GitHub Issues:**

### DIFF
--- a/pennylane_sf/gaussian.py
+++ b/pennylane_sf/gaussian.py
@@ -55,6 +55,7 @@ class StrawberryFieldsGaussian(StrawberryFieldsSimulator):
         wires (int): the number of modes to initialize the device in
         analytic (bool): indicates if the device should calculate expectations
             and variances analytically
+        cutoff_dim (int): Fock-space truncation dimension
         shots (int): Number of circuit evaluations/random samples used
             to estimate expectation values of observables. If ``analytic=True``,
             this setting is ignored.
@@ -93,6 +94,10 @@ class StrawberryFieldsGaussian(StrawberryFieldsSimulator):
     }
 
     _circuits = {}
+
+    def __init__(self, wires, *, analytic=True, cutoff_dim=10, shots=1000, hbar=2):
+        super().__init__(wires, analytic=analytic, shots=shots, hbar=hbar)
+        self.cutoff = cutoff_dim
 
     def pre_measure(self):
         self.eng = sf.Engine("gaussian")
--- a/pennylane_sf/simulator.py
+++ b/pennylane_sf/simulator.py
@@ -36,11 +36,13 @@ Code details
 ~~~~~~~~~~~~
 """
 import abc
+from collections import OrderedDict
 
 import numpy as np
 
 from pennylane import Device
 import strawberryfields as sf
+from strawberryfields.backends.states import BaseFockState, BaseGaussianState
 
 from ._version import __version__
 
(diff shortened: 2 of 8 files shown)


pennylane_sf/simulator.py
```diff
@@ -182,3 +184,46 @@ def observables(self):
...
+                marginal probabilities for. Wires not provided
+                are traced out of the system.
+
+        Returns:
+            OrderedDict[tuple, float]: Dictionary mapping a tuple representing the state
+            to the resulting probability. The dictionary should be sorted such that the
+            state tuples are in lexicographical order.
+        """
+        N = len(wires)
+        cutoff = getattr(self, "cutoff", 10)
```
trbromley, line 203: Do we need to add the `cutoff` attribute to `StrawberryFieldsGaussian` with a default?

josh146: Yep we do!

josh146: Just haven't had a chance yet haha

requirements.txt
```diff
@@ -1,2 +1,2 @@
-strawberryfields>=0.11.2
-pennylane>=0.7
+git+https://github.com/XanaduAI/strawberryfields.git
+git+https://github.com/XanaduAI/pennylane.git@probs_cv_cutoff
```
antalszava, line 2: temporary till the content of `probs_cv_cutoff` is incorporated in PL (this allows the `cutoff` value to be used when returning probabilities)

tests/test_fock.py
```diff
@@ -653,24 +761,3 @@ def circuit(n, a):
         gradF = circuit.jacobian([n, a], method="F")
         expected = np.array([2 * a ** 2 + 2 * n + 1, 2 * a * (2 * n + 1)])
         assert np.allclose(gradF, expected, atol=tol, rtol=0)
-
-    def test_tensor_number_operator_raises(self, tol):
```
antalszava, line 657: These tests were failing due to changes in SF

tests/test_fock.py
```diff
@@ -625,6 +627,112 @@ def circuit(n, a):
...
+        t3 = (a*np.exp(1j*phi))**(1-n)
+        t4 = np.sqrt((a*np.exp(1j*phi))**(2*n))
+        t5 = n
+        denom = fac(n)
+        partial_phi = t1*t2*t3*t4*t5/denom
+
+        params = [a, phi]
+        res_F = circuit.jacobian(params, method="F")
+        assert np.allclose(res_F[:,0], diff_a, atol=tol, rtol=0)
+        #assert np.allclose(res_F[:,1], partial_phi, atol=tol, rtol=0)
```
antalszava, line 734: @josh146 something funky seems to be going on here :thinking:

josh146: Have pushed a commit to fix the analytic formula here :)

tests/test_fock.py
```diff
@@ -625,6 +628,185 @@ def circuit(n, a):
...
+
+        @qml.qnode(dev)
+        def circuit(a, phi):
+            qml.Displacement(a, phi, wires=0)
+            return qml.probs(wires=[0])
+
+        a = 0.4
+        phi = -0.12
+        with pytest.raises(ValueError, match="Analytic gradients are not supported for circuits extracting Fock state probabilities"):
+            res_F = circuit.jacobian([a, phi], wrt={0}, method="A").flat
```
antalszava, line 808: This was added as per the error featured in https://github.com/XanaduAI/pennylane/pull/699

josh146:
@antalszava, have fixed up the squeezing test, and added `cutoff_dim` kwarg to the Gaussian device!

pennylane_sf/gaussian.py
```diff
@@ -55,6 +55,7 @@ class StrawberryFieldsGaussian(StrawberryFieldsSimulator):
         wires (int): the number of modes to initialize the device in
         analytic (bool): indicates if the device should calculate expectations
             and variances analytically
+        cutoff_dim (int): Fock-space truncation dimension
```
antalszava, line 58: :confetti_ball:

PR #40, opened 2020-06-22
GBS Variational

trbromley:
Use the following for prototyping:

```python
import pennylane as qml
import numpy as np
from pennylane_sf.vgbs import GraphTrain

cutoff = 5

dev = qml.device("strawberryfields.vgbs", wires=2, cutoff_dim=cutoff)

A = np.ones((2, 2))
n_mean = 4

@qml.qnode(dev)
def quantum_function(x):
    GraphTrain(x, A, n_mean, wires=range(2))
    return qml.probs(wires=[0, 1])

probs = quantum_function(np.ones(2))
print(probs)

dprobs = qml.jacobian(quantum_function, argnum=0)  # doesn't yet work
dprobs(np.ones(2))
```


trbromley:
Next step is to define the `jacobian` method for the device, although I'm not sure on what the signature should be there :thinking:

pennylane_sf/vgbs.py
```diff
@@ -0,0 +1,142 @@
...
+# import gates
+from strawberryfields.ops import Dgate, GraphEmbed, MeasureFock
+
+from .expectations import identity
+from .simulator import StrawberryFieldsSimulator
+
+
+class GraphTrain(qml.operation.CVOperation):
+    """TODO"""
+    do_check_domain = False # turn off parameter domain checking
```
josh146, line 52: By setting this class attribute, we can turn off the operation parameter validation 🙂 Avoids having to overwrite `__init__`, and allows each parameter to be a different type.

pennylane_sf/vgbs.py
```diff
@@ -0,0 +1,142 @@
...
+        WAW = W @ A @ W
+
+        singular_values = np.linalg.svd(WAW, compute_uv=False)
+        n_mean_WAW = np.sum(singular_values ** 2 / (1 - singular_values ** 2))
+
+        op = self._operation_map[operation](WAW, mean_photon_per_mode=n_mean_WAW / len(A))
+        op | [self.q[i] for i in wires] #pylint: disable=pointless-statement
+
+        if not self.analytic:
+            MeasureFock() | [self.q[i] for i in wires]
```
josh146, line 104: I added in support for 'non-analytic' mode (simply adding Fock measurements and specifying `shots=...` during engine run).

However, I have a question --- why use the SF frontend at all? For non-analytic mode, we could:

1. Simply use the `sf.apps.train` module, or

2. Directly call `thewalrus` directly.

Option (1) probably makes the most sense?

pennylane_sf/vgbs.py
```diff
@@ -0,0 +1,142 @@
...
+            # compute the fock probabilities analytically
+            # from the state representation
+            return super().probability(wires=wires)
+
+        # compute the fock probabilities from samples
+        N = len(wires) if wires is not None else len(self.num_wires)
+        probs = all_fock_probs_pnr(self.samples)
+        ind = np.indices([self.cutoff] * N).reshape(N, -1).T
+        probs = OrderedDict((tuple(k), v) for k, v in zip(ind, probs))
+        return probs
```
josh146, line 131: Estimating the probability from samples in non-analytic mode.

pennylane_sf/vgbs.py
```diff
@@ -0,0 +1,142 @@
...
+    def jacobian(self, operations, observables, variable_deps):
+        # NOTE: potentially could provide caching here, to avoid recomputing
+        # samples/probabilities.
+        self.reset()
+        prob = np.squeeze(self.execute(operations, observables, parameters=variable_deps))
+
+        # Compute and return the jacobian here.
+        # returned jacobian matrix should be of size (output_length, num_params)
+        jac = np.zeros([len(prob), len(self.weights)])
+        return jac
```
josh146, line 142: @trbromley, I've added scaffolding here for the jacobian method, let me know if you have any questions! I wasn't entirely sure of how to calculate the jacobian in this case.

You can make any of the variables in `apply()` attributes if you need to re-use them here.

pennylane_sf/vgbs.py
trbromley: Excellent, looks much better!

trbromley:
This can be used to prototype the `Cost` observable:
```python
import pennylane as qml
from pennylane import numpy as np
from pennylane_sf.vgbs import GraphTrain, Cost

cutoff = 5

dev = qml.device("strawberryfields.vgbs", wires=2, analytic=True, cutoff_dim=cutoff)
# dev = qml.device("strawberryfields.vgbs", wires=2, analytic=False, shots=100, cutoff_dim=cutoff)

def h(x):
    return np.sum(x)

A = np.ones((2, 2), requires_grad=False)
n_mean = np.array(4, requires_grad=False)

@qml.qnode(dev)
def quantum_function(x):
    GraphTrain(x, A, n_mean, wires=range(2))
    return qml.expval(Cost(h, wires=[0, 1]))
#     return qml.probs(wires=[0, 1])

w = np.ones(2, requires_grad=True)
expval = quantum_function(w)
print(expval)

dprobs = qml.jacobian(quantum_function, argnum=0)  # works now
grad = dprobs(np.ones(2))
print(grad)
```

trbromley:
Closing this PR as it was mainly for prototyping.
Implementation begins in https://github.com/PennyLaneAI/pennylane-sf/pull/47.

PR #42, opened 2020-07-05
Adding support for variance of the number operator

antalszava:
**Context:**
A recent addition in Strawberry Fields (https://github.com/XanaduAI/strawberryfields/pull/407) allowed support for extracting variances of the number operator for state objects. It also allowed for specifying more than 2 modes when the number expectation is returned using the `BaseGaussianState`.

With this change, the `state.number_expectation` method returns a tuple of `(expval, var)`. For compatibility reasons, an adjustment was needed in #39, which also enabled calculating variances of the number operator. Adding complete support for this feature, however, was out of scope for #39.

Adds test cases for features allowed by the changes to the SF `BaseGaussianState` class which include cases for:
* returning variances of the number operator and
* specifying more than 2 modes when returning the expectation value of the number operator using the Gaussian backend.

**Benefits:**
The new features from SF are tested in the plugin.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A


tests/test_fock.py
```diff
@@ -831,3 +831,143 @@ def circuit(n, a):
...
+        # differentiate with respect to r of the first displacement operation (rd0)
+        gradF = circuit.jacobian([pars], wrt={2}, method="F")
+        expected_gradient = pd_dr(*pars)
+        assert np.allclose(gradF, expected_gradient, atol=tol, rtol=0)
+
+        # differentiate with respect to r of the second squeezing operation (rd1)
+        gradF = circuit.jacobian([pars], wrt={6}, method="F")
+
+        expected_gradient = pd_dr(*reverted_pars)
+        assert np.allclose(gradF, expected_gradient, atol=tol, rtol=0)
```
josh146, line 973: Nice test!

tests/test_gaussian.py
```diff
@@ -73,10 +73,13 @@ def SF_gate_reference(sf_op, wires, *args):
 def SF_expectation_reference(sf_expectation, wires, *args):
     """SF reference circuit for expectation tests"""
     eng = sf.Engine("gaussian")
-    prog = sf.Program(2)
+
+    # Allows returning the variance of tensor number for 3 modes
+    num_wires = 2 if len(wires) <=2 else 3
```
josh146, line 78: This line seems a bit unclear 🤔 Should the number of wires always correspond to `len(wires)`?

tests/test_fock.py
co9olguy: I wonder if this should be split up into 3-4 separate tests, for better readability?

tests/test_gaussian.py
```diff
@@ -825,3 +828,141 @@ def circuit(n, a):
...
+        # differentiate with respect to r of the first displacement operation (rd0)
+        gradF = circuit.jacobian([pars], wrt={2}, method="F")
+        expected_gradient = pd_dr(*pars)
+        assert np.allclose(gradF, expected_gradient, atol=tol, rtol=0)
+
+        # differentiate with respect to r of the second squeezing operation (rd1)
+        gradF = circuit.jacobian([pars], wrt={6}, method="F")
+
+        expected_gradient = pd_dr(*reverted_pars)
+        assert np.allclose(gradF, expected_gradient, atol=tol, rtol=0)
```
co9olguy, line 968: same comment as above

tests/test_fock.py
antalszava: That's a good point! Was also thinking about it :thinking: Going for that would include having a longer part of code included in each test cases.

The main parts of these test cases could be factored out into fixtures before a new `TestVarianceDisplacedSqueezed` class:
```python
dev_list = [
            qml.device("strawberryfields.fock", wires=2, cutoff_dim=10),
            qml.device("strawberryfields.gaussian", wires=2)
            ]

@pytest.fixture(scope="function")
def disp_sq_circuit(dev):

    @qml.qnode(dev)
    def circuit(pars):
        qml.Squeezing(pars[0], pars[1], wires=0)
        qml.Displacement(pars[2], pars[3], wires=0)
        qml.Squeezing(pars[4], pars[5], wires=1)
        qml.Displacement(pars[6], pars[7], wires=1)
        return qml.var(qml.TensorN(wires=[0, 1]))

    return circuit

@pytest.fixture(scope="function")
def pars():
    """Parameters for the displaced squeezed circuit"""

    # Parameters to operations on the first mode
    # Squeezing
    rs0, phis0 = 0.1, 0.1

    # Displacement
    rd0, phid0 = 0.316277, 0.32175055
    alpha0 = rd0 * np.exp(1j*phid0)
    first_pars = np.array([rs0, phis0, rd0, phid0])

    # Parameters to operations on the second mode
    # Squeezing
    rs1, phis1 = 0.1, 0.15

    # Displacement
    rd1, phid1 = 0.14142136, 0.78539816
    alpha1 = rd1 * np.exp(1j*phid1)
    second_pars = np.array([rs1, phis1, rd1, phid1])

    return np.concatenate([first_pars, second_pars])
```

And each test case would then use the `disp_sq_circuit` and `pars` fixtures.

This would slightly differ from having explicit parts of the code in each test case (and could potentially end up in `conftest.py`). Would this be something that we'd like to go for or would we like to keep the tests entirely explicit?

tests/test_gaussian.py
antalszava: Valid point! Certain tests have `wires=[0]` (e.g. `test_one_mode_single_real_parameter_gates`) and use `SF_gate_reference`. For those this would be failing. Would it be better to have two separate `SF_gate_reference` definitions?

tests/test_fock.py
co9olguy: I was more asking about whether it makes sense to have so many different `assert`s in one function rather than having separate functions for each thing that is being tested.

Personally, when I read tests, I read them individually, so I don't even really notice code duplication. Fixtures are ok, as long as a user can easily find them and copy/paste them into their code

tests/test_gaussian.py
josh146: Ah, I see. I think it would be better to simply have `num_wires = len(wires)` here, and then below we can do

```python
with prog.context as q:
    sf.ops.Dgate(0.1) | q[0]
    if num_wires > 1:
            sf.ops.S2gate(0.1) | (q[0], q[1])
```

Everything should continue to work, but the logic is more readable

antalszava: One thing to note is that we'd like to create `Program(2)` even if `wires=[0]` (`wires` is used for `sf_expectation(state, wires, args)`).

Added a new `num_wires` parameter to the signature of `SF_expectation_reference`.

tests/test_fock.py
antalszava: makes sense, thank you! Broke it up into separate tests

antalszava:
@co9olguy @josh146 thank you for the comments!

* the tests are now included in a `test_shared.py`
* broke up the displaced squeezed test case into smaller ones
* adjusted the SF reference function for better readability

PR #47, opened 2020-08-04
Add GBS device with trainable probabilities

trbromley:
This PR adds a trainable GBS device that gives access to the GBS probability distribution and its derivative using the WAW encoding.

You can now do:
```python
import pennylane as qml
from pennylane_sf.ops import ParamGraphEmbed
from pennylane import numpy as np

dev = qml.device('strawberryfields.gbs', wires=4, cutoff_dim=4)

@qml.qnode(dev)
def quantum_function(x):
    ParamGraphEmbed(x, A, n_mean, wires=range(4))
    return qml.probs(wires=range(4))

A = np.array([[0.0, 1.0, 1.0, 1.0], [1.0, 0.0, 1.0, 0.0], [1.0, 1.0, 0.0, 0.0], [1.0, 0.0, 0.0, 0.0]])
n_mean = 2.5

d_quantum_function = qml.jacobian(quantum_function)

x = 0.9 * np.ones(4)
quantum_function(x)
d_quantum_function(x)
```
This can also be combined with PL optimizers:
```python
target = np.zeros(4 ** 4)  # all zero photons prob distribution
target[0] = 1

def all_zeros(x):
    p = quantum_function(x)
    return np.sum((p - target) ** 2)

opt = qml.GradientDescentOptimizer(0.3)

for i in range(30):
    x = opt.step(all_zeros, x)
```

Remaining tasks for future PRs:
- Add post-processing functions such as `kl_divergence()`
- Add reparametrization trick to speed up calculation of probabilities


trbromley:
[ch570]

pennylane_sf/gbs.py
```diff
@@ -0,0 +1,246 @@
...
+        wires (int): the number of modes to initialize the device in
+        analytic (bool): indicates if the device should calculate expectations
+            and variances analytically
+        cutoff_dim (int): Fock-space truncation dimension
+        backend (str): name of the remote backend to be used
+        shots (int): Number of circuit evaluations/random samples used
+            to estimate expectation values of observables. If ``analytic=True``,
+            this setting is ignored.
+        hbar (float): the convention chosen in the canonical commutation
+            relation :math:`[x, p] = i \hbar`
```
josh146, line 66: Does `hbar` have any effect with this device? Since the adjacency matrix is in the `a^\dagger` basis, and the output is always Fock probabilities

pennylane_sf/gbs.py
```diff
@@ -0,0 +1,246 @@
...
+    r"""StrawberryFields variational GBS device for PennyLane.
+
+    This device provides a method to embed variational parameters into GBS such that the analytic
+    gradient of the probability distribution is accessible.
+
+    Args:
+        wires (int): the number of modes to initialize the device in
+        analytic (bool): indicates if the device should calculate expectations
+            and variances analytically
+        cutoff_dim (int): Fock-space truncation dimension
```
josh146, line 60: 'Fock-space truncation dimension' sounds like an awesome band name or studio album

pennylane_sf/gbs.py
```diff
@@ -0,0 +1,246 @@
...
+            this setting is ignored.
+        hbar (float): the convention chosen in the canonical commutation
+            relation :math:`[x, p] = i \hbar`
+    """
+    name = "Strawberry Fields variational GBS PennyLane plugin"
+    short_name = "strawberryfields.gbs"
+
+    _operation_map = {
+        "ParamGraphEmbed": GraphEmbed,
+    }
```
josh146, line 73: Lol is it even worth having an operation map in this case? I suppose `StrawberryFieldsSimulator` forces you to

pennylane_sf/gbs.py
```diff
@@ -0,0 +1,246 @@
...
+            raise ValueError("Only the Gaussian backend is supported in non-analytic mode.")
+
+        super().__init__(wires, analytic=analytic, shots=shots, hbar=hbar)
+        self.cutoff = cutoff_dim
+        self.backend = backend
+        self._params = None
+        self._WAW = None
+
+    @staticmethod
+    def _calculate_WAW(params, A, n_mean):
```
josh146, line 94: Is it worth making this method public? Sounds like it could be quite useful.

josh146: Especially if you are curious in extracting the graph that was actually embedded.

pennylane_sf/gbs.py
```diff
@@ -0,0 +1,246 @@
...
+        ``n_mean``.
+
+        Args:
+            params (array[float]): variable parameters
+            A (array[float]): adjacency matrix
+            n_mean (float): mean number of photons
+
+        Returns:
+            array[float]: the :math`WAW` matrix
+        """
```
josh146, line 107: It already has the docstring you would expect from a public method as well :)

pennylane_sf/gbs.py
```diff
@@ -0,0 +1,246 @@
...
+        A *= rescale(A, n_mean)
+        W = np.diag(np.sqrt(params))
+        return W @ A @ W
+
+    @staticmethod
+    def _calculate_n_mean(A):
```
josh146, line 113: This could also be quite useful to be made public...

pennylane_sf/gbs.py
```diff
@@ -0,0 +1,246 @@
...
+        singular_values = np.linalg.svd(A, compute_uv=False)
+        return np.sum(singular_values ** 2 / (1 - singular_values ** 2))
+
+    # pylint: disable=pointless-statement,expression-not-assigned
+    def apply(self, operation, wires, par):
+        self._params, A, _ = par
+
+        if len(self._params) != self.num_wires:
+            raise ValueError(
+                "The number of variable parameters must be equal to the total number " "of wires."
```
josh146, line 131:
```suggestion
                "The number of variable parameters must be equal to the total number of wires."
```

pennylane_sf/gbs.py
```diff
@@ -0,0 +1,246 @@
...
+        """Calculate the mean number of photons for an adjacency matrix encoded into GBS.
+
+        Args:
+            A (array[float]): adjacency matrix
+
+        Returns:
+            float: mean number of photons
+        """
+        singular_values = np.linalg.svd(A, compute_uv=False)
+        return np.sum(singular_values ** 2 / (1 - singular_values ** 2))
```
josh146, line 123: Could this logic ever fail/return incorrect values?

pennylane_sf/gbs.py
```diff
@@ -0,0 +1,246 @@
...
+
+        if len(self._params) != self.num_wires:
+            raise ValueError(
+                "The number of variable parameters must be equal to the total number " "of wires."
+            )
+
+        self._WAW = self._calculate_WAW(*par)
+        n_mean_WAW = self._calculate_n_mean(self._WAW)
+
+        op = self._operation_map[operation](self._WAW, mean_photon_per_mode=n_mean_WAW / len(A))
```
josh146, line 137: I'm almost wondering if the code would be clearer simply doing
```suggestion
        op = GraphEmbed(self._WAW, mean_photon_per_mode=n_mean_WAW / len(A))
```

josh146: Another thought - is there any check anywhere to make sure that only _one_ `ParamGraphEmbed` appears in the circuit?

pennylane_sf/gbs.py
```diff
@@ -0,0 +1,246 @@
...
+            return super().probability(wires=wires)
+
+        N = len(wires)
+        samples = np.take(self.samples, self.wires.indices(wires), axis=1)
+
+        probs = all_fock_probs_pnr(samples)
+        ind = np.indices([self.cutoff] * N).reshape(N, -1).T
+
+        probs = OrderedDict((tuple(i), probs[tuple(i)]) for i in ind)
+        return probs
```
josh146, line 169: @antalszava this is another thing to add to the device ADR - standardize the form of returned probabilities!

pennylane_sf/gbs.py
```diff
@@ -0,0 +1,246 @@
...
+            return jac
+
+        jac = jac.reshape([self.cutoff] * self.num_wires + [self.num_wires])
+
+        all_indices = set(range(self.num_wires))
+        requested_indices = set(self.wires.indices(requested_wires))
+        trace_over_indices = all_indices - requested_indices
+
+        traced_jac = np.sum(jac, axis=tuple(trace_over_indices))
+        traced_jac = traced_jac.reshape(-1, self.num_wires)
```
josh146, line 203: What's happening here? Are you computing the jacobian over all wires, and then summing over rows/cols representing wires not measured?

pennylane_sf/ops.py
```diff
@@ -0,0 +1,60 @@
...
+        A (array): initial adjacency matrix
+        n_mean (float): initial mean number of photons
+        wires (Sequence[int] or int): the wire(s) the operation acts on
+    """
+
+    do_check_domain = False
+
+    num_params = 3
+    num_wires = AllWires
+    par_domain = "A"
```
josh146, line 60: Did there end up being any issues with the par domain?

tests/test_gbs.py
```diff
@@ -0,0 +1,608 @@
...
+    [
+        [
+            2.2071e00,
+            1.2071e00,
+            1.2071e00,
+            1.2071e00,
+            -2.6021e-18,
+            -1.3878e-16,
+            6.9323e-17,
+            -7.8505e-17,
```
josh146, line 40: Oh, black :(

If you want this to be formatted more like a matrix, feel free to use `#fmt: off` and `fmt: on` to turn black on and off around this array.

doc/devices/gbs.rst
```diff
@@ -0,0 +1,160 @@
...
+.. math::
+
+    \partial_{\mathbf{w}} P(\mathbf{n}, \mathbf{w}) = \frac{\mathbf{n} - \langle\mathbf{n}\rangle}{\mathbf{w}}P(\mathbf{n}, \mathbf{w}),,
+
+where :math:`\mathbf{n}` is a sample given by counting the number of photons observed in each mode.
+
+Device options
+~~~~~~~~~~~~~~
+
+The GBS device accepts additional arguments beyond the PennyLane default device arguments.
```
josh146, line 113: Seems to be missing the `backend` option

tests/test_gbs.py
```diff
@@ -0,0 +1,608 @@
...
+
+        @qml.qnode(dev)
+        def vgbs(params):
+            ParamGraphEmbed(params, A, 1, wires=wires)
+            return qml.probs(wires=[wires[subset_wires[0]], wires[subset_wires[1]]])
+
+        d_vgbs = qml.jacobian(vgbs, argnum=0)
+        dp = d_vgbs(params)
+
+        assert np.allclose(dp, jac_reduced[subset_wires])
```
josh146, line 608: Really great tests!

doc/devices/gbs.rst
```diff
@@ -0,0 +1,160 @@
+The GBS device
+==============
+
+The GBS device gives access to a near-term model of photonic quantum computing called `Gaussian
+boson sampling <https://strawberryfields.ai/photonics/concepts/gbs.html>`__. This model can
+encode graphs and other quantities that can be represented by a symmetric matrix :math:`A`, with
```
ixfoduap, line 6: Maybe good here to also mention the mean photon number. Otherwise it appears as a bit of a surprise in line 44

doc/devices/gbs.rst
```diff
@@ -0,0 +1,160 @@
...
+Hence, the ``quantum_function`` used to construct the QNode must follow a restricted pattern:
+
+.. code-block:: python
+
+    from pennylane_sf.ops import ParamGraphEmbed
+
+    @qml.qnode(dev)
+    def quantum_function(x):
+        ParamGraphEmbed(x, A, n_mean, wires=range(4))
+        return qml.probs(wires=range(4))
```
ixfoduap, line 45: More generally, what do we expect a VGBS qnode to return? Probabilities will scale badly for devices with more wires. My first thought is to return samples and always specify `shots` when defining the device

doc/devices/gbs.rst
```diff
@@ -0,0 +1,160 @@
...
+The GBS device can be trained by varying an initial matrix :math:`A` according to some trainable
+parameters :math:`\mathbf{w}`. One way to include trainable parameters is to transform :math:`A`
+according to
+
+.. math::
+
+    A \rightarrow WAW,
+
+where :math:`W` is a diagonal matrix with values given by :math:`\sqrt{\mathbf{w}}`. Using this
+approach, a `recent paper <https://arxiv.org/abs/2004.04770>`__ has shown how to calculate the
```
ixfoduap, line 101: The paper has been published, so we may consider linking there https://journals.aps.org/pra/abstract/10.1103/PhysRevA.102.012417

doc/devices/gbs.rst
ixfoduap: This should really only be simulated in the Gaussian backend though

pennylane_sf/gbs.py
ixfoduap: I'm not sure, but it may have an effect in how the covariance matrix is obtained from `A` and `n_mean`

pennylane_sf/gbs.py
ixfoduap: I think that the appearance of `Graph` here may be a bit mysterious since we have only been discussing a matrix A. There has to be a better name, but still, how about `MatrixEmbed`?

pennylane_sf/gbs.py
```diff
@@ -0,0 +1,246 @@
...
+        "Identity": identity,
+    }
+
+    _circuits = {}
+
+    _capabilities = {"model": "cv", "provides_jacobian": True}
+
+    def __init__(self, wires, *, analytic=True, cutoff_dim, backend="gaussian", shots=1000, hbar=2):
+        if not analytic and backend != "gaussian":
+            raise ValueError("Only the Gaussian backend is supported in non-analytic mode.")
```
ixfoduap, line 85: What is the reason to support other backends when `analytic=True`? Computing probabilities should also be much faster using the tricks from the walrus

pennylane_sf/gbs.py
ixfoduap: The idea is that A is only a valid matrix to be encoded if its singular values are smaller than 1. But you're right, if there is a singular value = 1 this will throw an error. Probably worth adding a check and a note in the docstring that the singular values of A have to be bounded this way

pennylane_sf/gbs.py
```diff
@@ -0,0 +1,246 @@
...
+        self._WAW = self._calculate_WAW(*par)
+        n_mean_WAW = self._calculate_n_mean(self._WAW)
+
+        op = self._operation_map[operation](self._WAW, mean_photon_per_mode=n_mean_WAW / len(A))
+        op | [self.q[wires.index(i)] for i in wires]
+
+        if not self.analytic:
+            MeasureFock() | [self.q[wires.index(i)] for i in wires]
+
+    def pre_measure(self):
```
ixfoduap, line 143: This is a note for myself: stopping here, to continue tomorrow

pennylane_sf/gbs.py
```diff
@@ -0,0 +1,246 @@
...
+        self.eng = sf.Engine(self.backend, backend_options={"cutoff_dim": self.cutoff})
+
+        if self.analytic:
+            results = self.eng.run(self.prog)
+        else:
+            # NOTE: currently, only the gaussian backend supports shots > 1
```
ixfoduap, line 149: Another reason why this device should perhaps be restricted to the Gaussian backend

josh146: I think the main reason for allowing multiple backends is for remote hardware support? @trbromley?

pennylane_sf/gbs.py
```diff
@@ -0,0 +1,246 @@
...
+
+    @staticmethod
+    def _calculate_WAW(params, A, n_mean):
+        """Calculate the :math:`WAW` matrix.
+
+        Rescales :math:`A` so that when encoded in GBS the mean photon number is equal to
+        ``n_mean``.
+
+        Args:
+            params (array[float]): variable parameters
```
ixfoduap, line 101: Just for my understanding, as opposed to the `embed` approach we used in SF, the idea with PL is that we can define the parameters `w` separately as functions of other parameters `theta` and the optimization will happen with respect to `theta`?

tests/test_gbs.py
```diff
@@ -0,0 +1,608 @@
...
+        """Test that the _calculate_WAW method calculates correctly when the input adjacency matrix is
+        already normalized to have a mean number of photons equal to 1."""
+        const = 2
+        A = 0.1767767 * np.ones((4, 4))
+        params = const * np.ones(4)
+        waw = StrawberryFieldsGBS._calculate_WAW(params, A, 1)
+        assert np.allclose(waw, const * A)
+
+    def test_calculate_n_mean(self):
+        """Test that the _calculate_n_mean calculates correctly"""
```
ixfoduap, line 306:
```suggestion
        """Test that _calculate_n_mean computes the mean photon number correctly"""
```

tests/test_gbs.py
ixfoduap: Yes! Very impressive and extremely thorough

pennylane_sf/gbs.py
```diff
@@ -0,0 +1,246 @@
...
+        samples = np.take(self.samples, self.wires.indices(wires), axis=1)
+
+        probs = all_fock_probs_pnr(samples)
+        ind = np.indices([self.cutoff] * N).reshape(N, -1).T
+
+        probs = OrderedDict((tuple(i), probs[tuple(i)]) for i in ind)
+        return probs
+
+    def jacobian(self, operations, observables, variable_deps):
+        """Calculate the Jacobian of the device.
```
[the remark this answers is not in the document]
trbromley: Done

pennylane_sf/gbs.py
trbromley: That's right @josh146! I also added some comments to help follow along @ixfoduap

pennylane_sf/ops.py
trbromley: Haven't encountered any :thinking:

tests/test_gbs.py
trbromley: Thanks, I was worried people would fall asleep with the number of tests :laughing:

doc/devices/gbs.rst
```diff
@@ -0,0 +1,160 @@
...
+
+Hence, the ``quantum_function`` used to construct the QNode must follow a restricted pattern:
+
+.. code-block:: python
+
+    from pennylane_sf.ops import ParamGraphEmbed
+
+    @qml.qnode(dev)
+    def quantum_function(x):
+        ParamGraphEmbed(x, A, n_mean, wires=range(4))
```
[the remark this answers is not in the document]
trbromley: Fixed with your nice suggestion below, thanks!

doc/devices/gbs.rst
trbromley: Returning samples is problematic for then accessing the gradient :thinking:

I think probabilities is good since you can set `analytic=False` and the number of shots, which will reconstruct the probability distribution and its derivatives from samples.

There could then be the question of whether we always make it `analytic=False`, but even though things scale badly I've had a better experience with ~4 modes using `analytic=True` than with a finite number of shots.

pennylane_sf/gbs.py
josh146: @trbromley it looks like all `StrawberryFieldsSimulator.probability()` does is set `sf.hbar = self.hbar`. We _could_ just set `self.hbar = 2` internally in the device as you suggest, but it sounds like best to keep it changeable by the user, as it could affect `n_mean`?

trbromley: I'm persuaded now - the choice of `hbar` is really only an internal parameter the user doesn't need to set it. I've removed it as an argument: https://github.com/PennyLaneAI/pennylane-sf/pull/47/files/1d4e04fa992f8c7782aac3ceb30e8773cc06073a..c9bc1fb805a615a3590ce099c4749c2fc91188ae

doc/devices/gbs.rst
trbromley: Changed to only using gaussian backend

pennylane_sf/gbs.py
trbromley: Thanks! Now only using gaussian backend.

pennylane_sf/gbs.py
trbromley: I've now changed to using only the `"gaussian"` backend.

It's right that on simulator one would only really want to use the `"gaussian"` backend - the `"fock"` backend is possible but would be slower, and also has the restriction on non-analytic mode.

There is an argument that we also might want to run on hardware or remote simulator. However, we aren't using `RemoteEngine` so it's anyway not possible in the current approach. It's also not clear to me how this would work with the `strawberryfields.remote` device - would we make a `strawberryfields.gbs.remote` device that inherits from `strawberryfields.gbs` and mixes in `strawberryfields.remote`? This looks like something to come back to.

doc/index.rst
```diff
@@ -43,6 +43,16 @@ PennyLane-SF provides three Strawberry Fields devices for PennyLane:
     <div style='clear:both'></div>
     </br>
 
+.. devicegalleryitem::
```
josh146, line 46: Don't forget to modify the above, where it says 'provides _three_ devices'!

josh146: >the two provided Strawberry Fields devices can be accessed

The 'two' can also be removed here

pennylane_sf/gbs.py
```diff
@@ -0,0 +1,258 @@
...
+
+Classes
+-------
+
+.. autosummary::
+   StrawberryFieldsGBS
+
+Code details
+~~~~~~~~~~~~
+"""
```
josh146, line 34: Worth double checking, but I don't think this is displayed in Sphinx anymore. I would remove all the Sphinx-specific markup, and simply have
```python
"""
This module contains the variational GBS device.

The Strawberry Fields variational GBS device is a simulator that provides a way to encode
variational parameters into GBS so that the gradient with respect to the output probability
distribution is accessible.
"""

File: pennylane_sf/gbs.py [hunk 28]
```diff
@@ -0,0 +1,258 @@
...
+
+from .expectations import identity
+from .simulator import StrawberryFieldsSimulator
+
+
+class StrawberryFieldsGBS(StrawberryFieldsSimulator):
+    r"""StrawberryFields variational GBS device for PennyLane.
+
+    This device provides a simulator that can embed variational parameters into GBS such that the
+    analytic gradient of the probability distribution is accessible.
```
josh146, line 55: I've been thinking about this lately; probably good to have a link back to the device page, in case a user ends up here by accident:

```suggestion
    r"""StrawberryFields variational GBS device for PennyLane.

    This device provides a simulator that can embed variational parameters into GBS such that the
    analytic gradient of the probability distribution is accessible.
    
    For more details see :doc:`/devices/gbs`.
```

pennylane_sf/gbs.py
josh146: I would buy it 💸

pennylane_sf/gbs.py
```diff
@@ -0,0 +1,258 @@
...
+        A *= rescale(A, n_mean)
+        W = np.diag(np.sqrt(params))
+        return W @ A @ W
+
+    @staticmethod
+    def calculate_n_mean(A):
+        """Calculates the mean number of photons for an adjacency matrix encoded into GBS.
+
+        Note that for ``A`` to be directly encoded into GBS, its singular values must not exceed
+        one.
```
josh146, line 111: Nitpick:
```suggestion
        .. warning::
        
            Note that for ``A`` to be directly encoded into GBS, its singular values must
            not exceed one.
```

josh146: Also, shouldn't this be 'its singular values must be less than 1'?

Another nitpick: This is likely something that could easily trip up users, since the only indication something went wrong would be a somewhat strange looking division by zero error. Is it worth having some validation to catch this, just to provide a more useful error message? Could even be as simple as

```python
singular_values = np.linalg.svd(A, compute_uv=False)

try:
    n_mean = np.sum(singular_values ** 2 / (1 - singular_values ** 2))
except DivideByZeroError:
    raise DeviceError("Singular values of matrix A must be less than 1")

return n_mean
```

Catching the divide by zero error is nice, as you don't have to worry about tolerances etc.

pennylane_sf/gbs.py
```diff
@@ -0,0 +1,258 @@
...
+        singular_values = np.linalg.svd(A, compute_uv=False)
+        return np.sum(singular_values ** 2 / (1 - singular_values ** 2))
+
+    # pylint: disable=missing-function-docstring
+    def execute(self, queue, observables, parameters=None, **kwargs):
+        parameters = parameters or {}
+        if len(queue) > 1:
+            raise ValueError(
+                "The StrawberryFieldsGBS device accepts a single application of ParamGraphEmbed"
+            )
```
josh146, line 128: nice :+1:

josh146: Nitpick:
```suggestion
        if len(queue) > 1:
            raise ValueError(
                "The StrawberryFieldsGBS device accepts only a single application of ParamGraphEmbed"
            )
```

pennylane_sf/gbs.py
```diff
@@ -0,0 +1,258 @@
...
+    def pre_measure(self):
+        self.eng = sf.Engine("gaussian", backend_options={"cutoff_dim": self.cutoff})
+
+        if self.analytic:
+            results = self.eng.run(self.prog)
+        else:
+            results = self.eng.run(self.prog, shots=self.shots)
+
+        self.state = results.state
+        self.samples = results.samples
```
josh146, line 158: Question: now that we are locked into _only_ using the Gaussian backend, why not go straight to The Walrus and simply use either `thewalrus.samples.hafnian_sample_graph` (non-analytic mode) or `thewalrus.quantum.probabilities` (analytic mode)? This should remove significant overhead.

The only advantage to going through the SF frontend that I can see is:

* The ability to hot-swap backends
* If you want to take advantage of compilation, e.g., if you are running on a hardware device (not yet supported).

pennylane_sf/gbs.py
```diff
@@ -0,0 +1,258 @@
...
+        self.reset()
+
+        requested_wires = observables[0].wires
+
+        # Create dummy observable to measure probabilities over all wires
+        obs_all_wires = qml.Identity(wires=self.wires)
+        obs_all_wires.return_type = Probability
+        prob = self.execute(operations, [obs_all_wires], parameters=variable_deps)[0]
+
+        jac = self._jacobian_all_wires(prob)
```
josh146, line 199: Question: is it possible to use the marginal probabilities to compute the jacobian over requested wires? Or is it more efficient to do it this way? Just wondering :)

pennylane_sf/gbs.py
```diff
@@ -0,0 +1,258 @@
...
+        """
+        jac = np.zeros([len(prob), self.num_wires])
+
+        n = len(self._WAW)
+        disp = np.zeros(2 * n)
+        cov = self.calculate_covariance(self._WAW, hbar=self.hbar)
+        mean_photons_by_mode = photon_number_mean_vector(disp, cov, hbar=self.hbar)
+
+        for i, s in enumerate(np.ndindex(*[self.cutoff] * self.num_wires)):
+            jac[i] = (s - mean_photons_by_mode) * prob[i] / self._params
```
josh146, line 238: Nitpick: I wonder if it is possible to vectorize this?
Something like
```python
basis_states = np.ndindex(*[self.cutoff] * self.num_wires)
indices = np.arange(len(basis_state))
jac = (basis_states - mean_photons_by_mode) * prob / self._params
```
(some modifications would be needed to the shapes to take into account `self._params` and `mean_photons_by_mode`, probably introducing an extra dimension).

pennylane_sf/gbs.py
```diff
@@ -0,0 +1,258 @@
...
+                :math:`[x, p] = i \hbar`
+
+        Returns:
+            array[float]: covariance matrix
+        """
+        n = len(A)
+        I = np.identity(2 * n)
+        o_mat = np.block([[np.zeros_like(A), np.conj(A)], [A, np.zeros_like(A)]])
+        cov = hbar * (np.linalg.inv(I - o_mat) - I / 2)
+        return cov
```
josh146, line 258: Weird that we don't have a function in The Walrus for this (@nquesada?) Using The Walrus, it looks like you would have to do

```python
>>> Covmat(gen_Qmat_from_graph(A, n_mean), hbar)
```

josh146: @thisac this would be nice to consider in The Walrus quantum refactor. Aside from organization, it would also be good to standardize naming for utility conversion functions :)

PR #48, opened 2020-08-06
[WIRES] Make plugin compatible with wires refactor

mariaschuld:
This PR updates the plugin to support the new wire management in PennyLane master.


pennylane_sf/expectations.py
```diff
@@ -42,13 +42,13 @@
 import pennylane.ops
```
mariaschuld, line 44: There were two options here: pass the "register" (`dev.wires`) to every expectation function and then do the extraction, or pass the `device_wires` directly. This does not work for `poly_xp`, but is still neater I think.

pennylane_sf/remote.py
```diff
@@ -150,11 +154,11 @@ def probability(self, wires=None):  # pylint: disable=missing-function-docstring
...
 
-        all_wires = np.arange(self.num_wires)
-        wires_to_trace_out = np.setdiff1d(all_wires, wires)
+        wires_to_trace_out = Wires.unique_wires([self.wires, wires])
+        device_wires_to_trace_out = self.map_wires(wires_to_trace_out)
 
-        if wires_to_trace_out.size > 0:
-            all_probs = np.sum(all_probs, axis=tuple(wires_to_trace_out))
+        if len(device_wires_to_trace_out) > 0:
+            all_probs = np.sum(all_probs, axis=device_wires_to_trace_out.labels)
```
mariaschuld, line 161: I think the size comparison was from a time where wires could be nested. This is now impossible.

mariaschuld:
I just realised, since SF does not work with the shared test suite, I should add some tests for the new wire logic here?

mariaschuld:
Latest version now contains the standard `TestWiresIntegration` test for both devices... all passing nicely.

tests/test_fock.py
```diff
@@ -574,6 +573,23 @@ def SF_gate_reference_trace(x, y):
         expected = SF_gate_reference_trace(r1, r2)
         assert np.allclose(circuit(r1, r2), expected, atol=tol, rtol=0)
 
+    def test_trace_subsystems(self, tol):
```
mariaschuld, line 576: Bit of a lame test, but I just want to check that the general pipeline works.

tests/test_fock.py
```diff
@@ -629,6 +645,23 @@ def circuit(n, a):
...
 
+    def test_polyxp_variance(self, tol):
+        """Tests that variance for PolyXP measurement works"""
+        dev = qml.device("strawberryfields.fock", wires=1, cutoff_dim=15)
+
+        @qml.qnode(dev)
+        def circuit(r, phi):
+            qml.Squeezing(r, 0, wires=0)
+            qml.Rotation(phi, wires=0)
+            return qml.var(qml.PolyXP(np.array([0, 1, 0]), wires=0))
```
mariaschuld, line 656: Idea here is to have `PolyXP` equal to `X` and recycle the expected data from previous tests.

pennylane_sf/expectations.py
```diff
@@ -206,7 +211,7 @@ def poly_xp(state, wires, params):
 
     # HACK, we need access to the Poly instance in order to expand the matrix!
     op = pennylane.ops.PolyXP(Q, wires=wires, do_queue=False)
-    Q = op.heisenberg_obs(state.num_modes)
+    Q = op.heisenberg_obs(all_wires)
```
antalszava, line 214: Saw the [comment related to why all wires need to be passed to `heisenberg_obs`](https://github.com/PennyLaneAI/pennylane/pull/666/files#r458117642), very thorough thinking there! :blush:

Still trying to put together the details, one thing that came to mind is, could the QNode extract the `wire_indices` and then `heisenberg_obs` would "blindly" operate using those indices? One advantage would be that there's less complexity for `heisenberg_obs` and an object that is mostly the caller to it (QNode) could take this burden. Even if it would work, perhaps not too pressing as such a shift could be nice with several use cases not just `heisenberg_obs` :thinking:

tests/test_wires.py
```diff
@@ -0,0 +1,86 @@
...
+    @pytest.mark.parametrize("circuit_factory", [make_simple_circuit_expval])
+    def test_wires_gaussian(self, circuit_factory, wires1, wires2, tol):
+        """Test that the expectation of the gaussian device is independent from the wire labels used."""
+        dev1 = qml.device("strawberryfields.gaussian", wires1)
+        dev2 = qml.device("strawberryfields.gaussian", wires2)
+
+        circuit1 = circuit_factory(dev1, wires1)
+        circuit2 = circuit_factory(dev2, wires2)
+
+        assert np.allclose(circuit1(), circuit2(), tol)
```
antalszava, line 86:
```suggestion
        assert np.allclose(circuit1(), circuit2(), tol)

```

pennylane_sf/remote.py
```diff
@@ -42,7 +44,9 @@ class StrawberryFieldsRemote(StrawberryFieldsSimulator):
     configuration file.
 
     Args:
-        wires (int): the number of modes to initialize the device in
+        wires (int, Iterable[Number, str]]): Number of subsystems accessible on the device,
+            or iterable that contains unique labels for the subsystems as numbers (i.e., ``[-1, 0, 2]``)
+            or strings (``['ancilla', 'q1', 'q2']``).
```
antalszava, line 49:
```suggestion
        wires (Iterable[Number, str]]): Iterable that contains unique labels for the subsystems as numbers (i.e., ``[-1, 0, 2]``)
            or strings (``['ancilla', 'q1', 'q2']``).
```
Passing custom wires will error; `wires` is not a valid argument at the moment (nor is `kwargs` a part of the signature for `__init__`).

This could be mended by:
* Adding a keyword-only `wires` parameter to the signature
* Making `wires` **only** accept an `Iterable`: previously `wires` was not added because the number of wires is inferred from SF by specifying a specific backend
* Checking that the number of unique labels specified equals the number of modes for the backend
* Adding testing that involves mocking out `sf.RemoteEngine` and instantiates `StrawberryFieldsRemote` with unique labels

pennylane_sf/remote.py
antalszava: Would be worth adding a test case on unique labels to see that this one comes around.

pennylane_sf/simulator.py
```diff
@@ -115,13 +119,23 @@ def expval(self, observable, wires, par):
...
         Returns:
             float: expectation value
         """
-        ex, var = self._observable_map[observable](self.state, wires, par)
+        # translate to consecutive wires used by device
+        device_wires = self.map_wires(wires)
+
+        # The different "expectation" functions require different inputs,
+        # which is at the moment solved by having dummy arguments.
+        # This one-size-fits all "observable_map" logic should be revised.
```
antalszava, line 133: Fair point!

pennylane_sf/simulator.py
```diff
@@ -198,6 +222,12 @@ def probability(self, wires=None):
             to the resulting probability. The dictionary should be sorted such that the
             state tuples are in lexicographical order.
         """
+        wires = wires or self.wires
```
antalszava, line 225: Just to make sure: by this point, it was already checked in the QNode that the labels make sense with regards to the labelling of the device, correct? :thinking:

tests/test_fock.py
antalszava: Would it be worth adding cases for unique wire labels?

pennylane_sf/expectations.py
mariaschuld: I'm not sure what is the best way to deal with ``operator`` functions that need to know how wires are arranged on a register. You are right, for ``heisenberg_obs`` we could reduce the number of arguments passed around by using wire indices directly. But in the long run I would almost prefer moving all those functions into the qnode or better, the device, since they are part of applying operators to circuits.

pennylane_sf/simulator.py
mariaschuld: Yes, line 368 in base.py checks this. And it would be picked up by ``map_wires`` as well, although with a more specific error message.

pennylane_sf/remote.py
mariaschuld: Great, thanks for picking this up! I will try to go the `Iterable` only route. Especially for the hardware labellings that show how the first half of the wires is related to the second could be nice.

pennylane_sf/expectations.py
mariaschuld: Actually, @josh146 what do you think about this?

josh146: I have to admit, I am not that up to date with the discussion, but my understanding is that the ordering comes into play due to the `heisenberg_expand` method, which takes the subset of wires and the _total_ number of wires, and tries to expand the operator to act on the full system (by inserting identities where required).

It definitely feels like logic that should be a part of the `CVQNode`.

pennylane_sf/remote.py
mariaschuld: I allowed for int wires to be consistent with other devices, but check that the integer matches the number of modes. Or would you prefer only allowing iterables? 

Even though the integer option is a dummy since only one value is allowed, it gives the user an informative message when doing it wrong...

pennylane_sf/remote.py
mariaschuld: I'm wondering, isn't that case tested in `test_mocked_probability`? Since `probs` is called with 2 out of 8 wires? Or do I understand your comment wrong?

mariaschuld:
Thanks @antalszava for picking op the remote input issue. Would this solution work for you?

pennylane_sf/remote.py
```diff
@@ -121,8 +144,9 @@ def sample(
     ):  # pylint: disable=unused-argument, missing-function-docstring
         if observable == "Identity":
             return np.ones(self.shots)
-        wires = np.array(wires)
-        selected_samples = self.samples[:, wires]
+        device_wires = self.map_wires(wires)
+
```
mariaschuld, line 148:
```suggestion
```

pennylane_sf/remote.py
antalszava: Oh, sorry for the wording! Was thinking of a small example where the device has a wires labeling defined, e.g. by defining ["a", "c", "d","e","f","g","h","i"] as wires. Then 2 of them could be passed to `qml.probs`. In `test_mocked_probability` the `wires` variable is not actually used when the device is being defined. :confused:

pennylane_sf/remote.py
```diff
@@ -87,12 +89,29 @@ class StrawberryFieldsRemote(StrawberryFieldsSimulator):
...
         eng = sf.RemoteEngine(self.backend)
 
-        # Infer the number of modes from the device specs
-        wires = eng.device_spec.modes
+        self.num_wires = eng.device_spec.modes
+
+        if wires is None:
+            # infer the number of modes from the device specs
+            # and use consecutive integer wire labels
+            wires = range(self.num_wires)
```
antalszava, line 101: :1st_place_medal:

tests/test_remote.py
```diff
@@ -99,6 +100,22 @@ def quantum_function():
...
+
+        with pytest.raises(ValueError, match="Device has a fixed number of"):
+            qml.device("strawberryfields.remote", wires=8, backend="X8", shots=10)
+
+        dev_iterable_wires = qml.device("strawberryfields.remote", wires=range(8), backend="X8", shots=10)
+        assert dev_iterable_wires.wires == Wires(range(8))
+
+        with pytest.raises(ValueError, match="Device has a fixed number of"):
+            qml.device("strawberryfields.remote", wires=range(9), backend="X8", shots=10)
+
```
antalszava, line 118: (Minor suggestion): could be worth splitting this up into 4 short tests

tests/test_wires.py
```diff
@@ -83,4 +136,33 @@ def test_wires_gaussian(self, circuit_factory, wires1, wires2, tol):
...
+        dev2 = qml.device("strawberryfields.remote", wires=wires2, backend="X8", shots=shots)
+
+        circuit1 = make_x8_circuit_expval(dev1, wires1)
+        circuit2 = make_x8_circuit_expval(dev2, wires2)
+
+        monkeypatch.setattr(
+            "pennylane_sf.remote.samples_expectation", lambda *args: expected_expval
+        )
+
+        assert np.allclose(circuit1(1.0, 0), circuit2(1.0, 0), tol)
```
antalszava, line 168: Awesome! :slightly_smiling_face:

pennylane_sf/remote.py
```diff
@@ -42,9 +44,13 @@ class StrawberryFieldsRemote(StrawberryFieldsSimulator):
     configuration file.
 
     Args:
-        wires (int): the number of modes to initialize the device in
         shots (int): number of circuit evaluations/random samples used to
             estimate expectation values of observables
+        wires (Iterable[Number, str]): Iterable that contains unique labels for the
+            modes as numbers or strings (i.e., ``['m1', ..., 'm4', 'n1',...,'n4']``).
+            The number of labels must match the number of modes accessible on the backend.
+            If not provided, modes are addressed as consecutive integers [0, 1, ...], and their number
```
antalszava, line 52:
```suggestion
            If not provided, modes are addressed as consecutive integers ``[0, 1, ...]``, and their number
```

pennylane_sf/remote.py
mariaschuld: Something like this test?

```    python
    
        shots = 10
        wires = ["a", "c", "d", "b", "f", "e", "h", "g"]

        monkeypatch.setattr("strawberryfields.RemoteEngine", MockEngine)
        device = qml.device("strawberryfields.remote", wires=wires, backend="X8", shots=shots)

        @qml.qnode(device)
        def circuit():
            return qml.probs(wires=["a"])

        expected = np.array([0., 0.4, 0.2, 0.1, 0.3])

        assert np.allclose(circuit(), expected)
```

I must say I don't really understand the expected output given ``MOCK_DEVICE``s samples

```
MOCK_SAMPLES = np.array(
    [
        [3, 4, 2, 3, 4, 3, 1, 0],
        [4, 3, 3, 2, 0, 3, 1, 4],
        [2, 1, 3, 3, 3, 2, 2, 4],
        [4, 1, 4, 4, 2, 3, 3, 0],
        [4, 2, 3, 3, 3, 0, 0, 4],
        [1, 2, 4, 4, 2, 0, 0, 4],
        [2, 3, 1, 2, 1, 0, 4, 1],
        [1, 2, 0, 1, 2, 3, 3, 0],
        [1, 2, 4, 0, 0, 4, 2, 4],
        [1, 0, 1, 1, 1, 3, 1, 0],
    ]
)
```

At the moment, the expected output is just retrieved from running the test and copying the actual result - but of course I want to understand it before merging...

mariaschuld: Do you see why the probability of the first wire is `[0., 0.4, 0.2, 0.1, 0.3]`? What are the 5 numbers here? Is there an implicit cutoff of 5?

mariaschuld: Feel free to merge if this added test makes sense to you, otherwise I'll get back to it tomorrow!

tests/test_wires.py
```diff
@@ -166,3 +166,20 @@ def test_wires_remote(self, wires1, wires2, tol, monkeypatch):
...
+        monkeypatch.setattr("strawberryfields.RemoteEngine", MockEngine)
+        device = qml.device("strawberryfields.remote", wires=wires, backend="X8", shots=shots)
+
+        @qml.qnode(device)
+        def circuit():
+            return qml.probs(wires=["a"])
+
+        expected = np.array([0., 0.4, 0.2, 0.1, 0.3])
+
+        assert np.allclose(circuit(), expected)
```
antalszava, line 185: 💯

PR #49, opened 2020-08-07
Update section on device arguments in remote.rst 

antalszava:
Updates the arguments section for the remote device. As such sections might easily change over time, might consider resorting to simple automatic documentation from sphinx.


antalszava:
@mariaschuld thanks for the note! Added a section for `wires` too, mostly depending on the docstring of the remote device. The tests would continue to fail until #48 is merged, guess this could be merged right after that.

PR #54, opened 2020-09-03
Use a fixed cutoff for remote engine

trbromley:
The cutoff dimension is primarily used in the `probability()` method of `StrawberryFieldsRemote`. Currently, it is set dynamically using the size of `all_probs`:
```python
all_probs = all_fock_probs_pnr(self.samples)
        
# Extract the cutoff value by checking the number of Fock states we
# obtained probabilities for
cutoff = all_probs.shape[0]
```
However, this plays havoc when trying to calculate the jacobian of the probability distribution and is not compatible with some of the assumptions in PL core:
- Without `self.cutoff` being set, the `grad` array that holds the result in `JacobianQNode.jacobian()` can be of the incorrect size. The `grad` array size is set by `self.output_dim` in `BaseQNode._check_circuit()`, which defaults to using a cutoff of 10 if not set. This becomes a problem when the probability distribution ends up with a cutoff of, e.g., 4 since at most 3 photons were detected.
- The `JacobianQNode._pd_finite_diff()` method calculates `(y - y0) / h`. Unfortunately, the size of `y` and `y0` can differ if `self.cutoff` is not uniform. For example, `y0` saw at most 3 photons in one mode, while `y` saw at most 4.

This PR solves the issue by setting a fixed cutoff dimension. Like other devices, this can be set by the user and has a default value of `5`. With this, we can now differentiate the probability distribution:
```python
import pennylane as qml

dev = qml.device('strawberryfields.remote', backend="X8", shots=1000, sf_token="...")  # cutoff defaults to 5

@qml.qnode(dev)
def quantum_function(theta, phi):
    qml.TwoModeSqueezing(1.0, 0.0, wires=[0,4])
    qml.Beamsplitter(theta, phi, wires=[0,1])
    qml.Beamsplitter(theta, phi, wires=[4,5])
    return qml.probs(wires=[0])

d_quantum_function = qml.jacobian(quantum_function, argnum=0)

d_quantum_function(0.5, 0.4)
```

One potential drawback is that the Autograd interface evaluates the jacobian using a number of calls to `probability()` that depends on `self.cutoff`, e.g., the above circuit evaluates `probability()` 16 times (with a cutoff of 5). Note that this doesn't happen using the TF interface, which evaluates 3 times. Hence, the user can create a big overhead if they set the cutoff too high and use the autograd interface. However, for now I think this is the best approach.


pennylane_sf/remote.py
```diff
@@ -44,14 +44,15 @@ class StrawberryFieldsRemote(StrawberryFieldsSimulator):
     configuration file.
 
     Args:
-        shots (int): number of circuit evaluations/random samples used to
-            estimate expectation values of observables
+        backend (str): name of the remote backend to be used
```
trbromley, line 47: (also rearranged order to match the function signature)

pennylane_sf/remote.py
```diff
@@ -159,17 +161,15 @@ def var(self, observable, wires, par):
...
     def probability(self, wires=None):  # pylint: disable=missing-function-docstring
-        all_probs = all_fock_probs_pnr(self.samples)
-
-        # Extract the cutoff value by checking the number of Fock states we
-        # obtained probabilities for
-        cutoff = all_probs.shape[0]
+        fock_probs = all_fock_probs_pnr(self.samples)
+        cutoff = fock_probs.shape[0]
+        diff = max(self.cutoff - cutoff, 0)
+        all_probs = np.pad(fock_probs, [(0, diff)] * self.num_wires)
```
josh146, line 167: Ah, so if the 'actual' cutoff is below the set cutoff, the fock probabilities are just padded with 0s?

tests/test_remote.py
```diff
@@ -330,13 +332,62 @@ def quantum_function(theta, phi):
...
 
-        # all_fock_probs uses a cutoff equal to 1 + the maximum number of
-        # photons detected (including no photons detected)
-        cutoff = 1 + np.max(MOCK_SAMPLES)
+        assert probs.shape == (dev.cutoff ** len(wires),)
+        assert np.allclose(probs, expected_probs, atol=tol)
+
+    def test_mocked_engine_run_all_fock_probs_larger_cutoff(self, monkeypatch, tol):
+        """Test that the probabilities are returned as expected in the special case that the
+        cutoff dimension is larger than the implied cutoff dimension of the samples."""
```
josh146, line 340: answers my question above!

josh146 (approved):
Looks good Tom! Good catch as well

pennylane_sf/remote.py
```diff
@@ -159,17 +161,15 @@ def var(self, observable, wires, par):
...
 
         if wires is None:
-
             all_probs = all_probs.flat
             N = self.num_wires
-            ind = np.indices([cutoff] * N).reshape(N, -1).T
+            ind = np.indices([self.cutoff] * N).reshape(N, -1).T
```
antalszava, line 172: What happens in the `cutoff > self.cutoff` case here? :thinking: Would some values be disregarded?

antalszava (approved):
Looks good to me, had one question. Thanks for catching this @trbromley! :confetti_ball:

pennylane_sf/remote.py
trbromley: Yes, we'd be ignoring some values. However, I think this is ok and in line with what one would expect when calculating probabilities with a fixed cutoff, i.e., sometimes you lose some data.

Although I do agree that in this case it could be a shame to throw away useful samples with a high number of photons when it comes from actual hardware. If the user really wants to keep all information, maybe they'd be better to stick with returning samples?

trbromley (commented):
@antalszava, you made me realise that it's a lot more subtle in the case where the device's cutoff is below the cutoff implied by samples! This has led to a bit of a rethink of how things are done and some changes to the tests (which I now believe to be more comprehensive, covering both where probabilities are returned on a subset and all of the modes).

I've tagged you both for re-review, apologies for the repetition but I think it's a big enough change from before.

Also, coverage is at 100%:
image

pennylane_sf/remote.py
```diff
@@ -161,20 +162,32 @@ def var(self, observable, wires, par):
...
-        wires_to_trace_out = Wires.unique_wires([self.wires, wires])
-        device_wires_to_trace_out = self.map_wires(wires_to_trace_out)
+            sl = []
+            for wire in device_wires:
+                if wire in device_wires_to_trace_out:
+                    sl.append(slice(None))
+                else:
+                    sl.append(slice(self.cutoff))
+
+            all_probs = fock_probs[sl]
```
antalszava: Just for my own understanding: previously thought that using `self.cutoff` in `ind = np.indices([self.cutoff] * N).reshape(N, -1).T` would already "remove" certain cutoff values (as not all basis state indices would be generated if `self.cutoff < cutoff`).

With this addition the relevant values are also explicitly removed. Was clarity the reason behind that or was there anything else as well?

antalszava (approved):
Looks good to me! :100:

pennylane_sf/remote.py
josh146: There might be subtle differences here between simulation and hardware (and not sure I fully understand the implications).

In the simulation case, we specify a truncated space where all computation takes place. For the Gaussian PNR samples, no information is discarded, but rather the samples might not be an accurate representation since we are always enforcing a probability of 1 across the truncated subspace.

In the hardware case, we are instead discarding additional info?

tests/test_remote.py
```diff
@@ -284,60 +322,118 @@ def quantum_function(theta, phi):
...
+            qml.Beamsplitter(theta, phi, wires=[4, 5])
+            return qml.probs(wires=wires)
+
+        with pytest.warns(UserWarning, match="Samples were generated where at least one mode"):
+            probs = quantum_function(1.0, 0)
+
+        # with dev.cutoff = 3, probs will be a 9-dimensional flat array. Since
+        # partial_probs is a 25-dimensional flat array, we first reshape to a 5x5 matrix and then
+        # take the first 3 elements on both axes, finally flattening
+        exp_probs = partial_probs.reshape((5, 5))[:3, :3]
```
josh146, line 435: Just double checking, but re-normalization does _not_ happen, right? I think that is the right call, since unnormalized probabilities can act as a warning to the user (in addition to the actual warning 😆 ) that their cutoff was too small.

josh146 (approved):

josh146:
Thanks Tom! If codecov is playing up, will merge this in once @antalszava and my comments are resolved

pennylane_sf/remote.py
trbromley: :thinking: yes, in the hardware case we can have samples with any number of photons without a cutoff. We then create `fock_probs`, which necessarily sums to 1 since it includes all of the probabilities. This is then shrunk down if the cutoff is below the cutoff inferred from `fock_probs`.

This is probably different to (a) samples from a simulator with a fixed cutoff and (b) calculating the exact probability distribution. Not really sure what the solution is though other than just being aware.

trbromley (commented):
> Just for my own understanding: previously thought that using self.cutoff in ind = np.indices([self.cutoff] * N).reshape(N, -1).T would already implicitly "remove" certain values from all_probs (as not all basis state indices would be generated if self.cutoff < cutoff).
>
>With this addition, the relevant values are also explicitly removed. Was clarity the reason behind that or was there anything else as well?

@antalszava, relying on just `ind` is not enough, I had a double check this morning. We loop over `zip(ind, all_probs)` to create the dictionary, so we also need to be careful about `all_probs`. So when `self.cutoff < cutoff`, if we leave `all_probs` to have `cutoff` dimensions, then it's flattened version will be in blocks of terms `[[cutoff], [cutoff], [cutoff], ...]`, but we need it to be in blocks of terms `[[self.cutoff], [self.cutoff], [self.cutoff], ...]`. It's hard to explain so if still not clear I can make an example. But in summary, it is not just a cosmetic/clarity-based change.

tests/test_remote.py
josh146: > But for me this is worse.

I agree! I like the current approach.

josh146 (approved):

antalszava (approved):

PR #52, opened 2020-09-01
[WIP] Adds reparametrization trick

trbromley:
WIP

PR #55, opened 2020-09-04
Adds reparametrization trick for samples in StrawberryFieldsGBS

trbromley:
This PR adds the ability for the `StrawberryFieldsGBS` device to use the reparametrization trick in sampling mode.

Now, if `analytic=False` (i.e., sampling mode) and `use_cache=True`, the device will generate samples from the input adjacency matrix `A` rather than `WAW`, and then reparametrize to get the probability distribution of `WAW`. On subsequent calls, the device will see that samples are already present and skip generating new samples. There is also the option for users to input pre-generated samples using the `samples` argument.

Note that it is also possible to use caching in analytic mode, i.e., saving the probability distribution from `A` and reparametrizing as above. This was explored in https://github.com/PennyLaneAI/pennylane-sf/pull/52, however it transpired to be _slower_ than simply recalculating the probability distribution. A lot of the overhead in reparametrization is to loop over all of the samples up to a cutoff (there are `cutoff ** modes` samples). In analytic mode, it is faster to call `all_fock_probs(cutoff)`, which is probably because we don't get into the regime where Hafnian calculations are tough. Instead, reparametrizing in sampling mode makes a lot more sense, since sampling takes a long time.


pennylane_sf/gbs.py
```diff
@@ -124,46 +134,110 @@ def execute(self, queue, observables, parameters=None, **kwargs):
...
-        ind = np.indices([self.cutoff] * N).reshape(N, -1).T
+        fock_probs = all_fock_probs_pnr(samples)
+        cutoff = fock_probs.shape[0]
+        diff = max(self.cutoff - cutoff, 0)
+        probs = np.pad(fock_probs, [(0, diff)] * len(wires))
+
+        if self.use_cache:
+            if len(wires) < self.num_wires:
+                raise ValueError("Caching is only supported when returning the probabilities on "
+                                 "all of the wires")
```
trbromley, line 237: This could feasibly be fixed to work with subsets of wires, but I realised having a go today that it's nontrivial and would take some work (especially adding tests).

Since I view this as more of an edge case anyway, I'd rather we simply raise an error here and see if we get requests from users for caching + subsets of wires.

pennylane_sf/gbs.py
```diff
@@ -124,46 +134,112 @@ def execute(self, queue, observables, parameters=None, **kwargs):
...
+        cutoff = fock_probs.shape[0]
+        diff = max(self.cutoff - cutoff, 0)
+        probs = np.pad(fock_probs, [(0, diff)] * len(wires))
+
+        if self.use_cache:
+            if len(wires) < self.num_wires:
+                raise ValueError(
+                    "Caching is only supported when returning the probabilities on "
+                    "all of the wires"
+                )
```
trbromley, line 239: This could feasibly be fixed to work with subsets of wires, but I realised having a go today that it's nontrivial and would take some work (especially adding tests).

Since I view this as more of an edge case anyway, I'd rather we simply raise an error here and see if we get requests from users for caching + subsets of wires.

trbromley:
[ch647]

doc/devices/gbs.rst
```diff
@@ -117,6 +117,23 @@ has shown how to calculate the derivative of the output GBS probability distribu
 
 where :math:`\mathbf{n}` is a sample given by counting the number of photons observed in each mode.
 
+Moreover, the probability :math:`P(\mathbf{n}, \mathbf{w})` of a sample :math:`\mathbf{n}`
+according to trainable parameters :math:`\mathbf{w}` can be calculated as:
+
+.. math::
+
+    P(\mathbf{n}, \mathbf{w}) = \frac{Z(A)}{Z(WAW)} \prod_{i=1}^{n} w_{i}^{n_{i}}
```
ixfoduap, line 125:
```suggestion
    P(\mathbf{n}, \mathbf{w}) = \frac{Z(A)}{Z(WAW)} \prod_{i=1}^{n} w_{i}^{n_{i}},
```

doc/devices/gbs.rst
```diff
@@ -117,6 +117,23 @@ has shown how to calculate the derivative of the output GBS probability distribu
...
+
+where :math:`Z` is a normalization coefficient giving the probability of detecting no photons.
```
ixfoduap, line 127:
```suggestion
where :math:`Z` is a normalization coefficient.
```
I'm not sure it's helpful to mention that connection

doc/devices/gbs.rst
```diff
@@ -117,6 +117,23 @@ has shown how to calculate the derivative of the output GBS probability distribu
...
+This means that the probability distribution and its derivative can be calculated directly from
+the probability distribution of :math:`A`. This observation is particularly useful for devices in
+non-analytic mode, allowing us to generate a reference set of samples from :math:`A` and rescale
+the resulting probability distribution to give :math:`P(\mathbf{n}, \mathbf{w})` for any choice of
+:math:`\mathbf{w}`.
+
+Caching samples from :math:`A` is possible by setting the ``use_cache=True`` argument. It is also
```
ixfoduap, line 134: Maybe it's my ignorance, but the meaning of "caching" is not clear to me, so maybe it's not for the average user either? Might be nice to add a short clarification here

ixfoduap (commented):
Will continue my review later :-)

doc/devices/gbs.rst
trbromley: Thanks, I've tried to improve a6c6ddc654a45a36986b97f64c4ca3f3bdb78d81

pennylane_sf/gbs.py
```diff
@@ -66,28 +75,29 @@ class StrawberryFieldsGBS(StrawberryFieldsSimulator):
...
+        self.use_cache = use_cache
+        self.samples = samples
+
         self._params = None
         self._WAW = None
+        self.Z_inv = None
 
     @staticmethod
-    def calculate_WAW(params, A, n_mean):
+    def calculate_WAW(params, A):
```
ixfoduap, line 91: Is this change for the purpose of this PR, or a more general change that is being included here for convenience? In other words, what's the reason for removing `n_mean` as an argument?

ixfoduap (commented):
Looks good to the best of my ability, but I have to admit I had a difficult time following the code closely.

My main suggestion is that it may be worthwhile to add a concrete example of caching and use of the reparametrization trick to the "usage" section of the docs. Right now, I don't think it is sufficiently clear to the users how to make use of this capability

trbromley (commented):
Thanks @ixfoduap, was useful to add an example!

doc/devices/gbs.rst
```diff
@@ -129,6 +166,15 @@ The GBS device accepts additional arguments beyond the PennyLane default device
 	The number of circuit evaluations/random samples used to estimate probabilities.
 	Only used when ``analytic=False``, otherwise probabilities are exact.
 
+``use_cache``
+    Indicates whether to use samples from previous evaluations to speed up calculation of the
```
antalszava, line 170:
```suggestion
    Indicates whether to use samples from previous evaluations to speed up the calculation of the
```

pennylane_sf/gbs.py
```diff
@@ -124,46 +134,112 @@ def execute(self, queue, observables, parameters=None, **kwargs):
...
+            op = GraphEmbed(self._WAW, mean_photon_per_mode=n_mean_WAW / len(A))
 
-        op = GraphEmbed(self._WAW, mean_photon_per_mode=n_mean_WAW / len(A))
         op | [self.q[wires.index(i)] for i in wires]
 
         if not self.analytic:
             MeasureFock() | [self.q[wires.index(i)] for i in wires]
 
+    def reset(self):
+        # There is no engine to reset in this case
```
antalszava, line 160: Wait, this assumes that the engine was created by passing the arguments `use_cache=True` and `samples=<some_samples_array>`. What if that was not the case, but for example, the `samples`  attribute was `None` and it gets set later?
```python
dev = qml.device('strawberryfields.gbs', wires=4, cutoff_dim=4,
                samples=None, use_cache=True, analytic=False)

# executing a circuit on dev, samples are stored as per `dev.pre_measure`
dev.reset()  # no reset mechanism, dev.eng still remains with samples, state, etc.
```

antalszava (approved):
Looks good to me! :100: Had one comment regarding the `reset`

pennylane_sf/gbs.py
trbromley: There are two options in non-analytic + caching mode, i.e.,
1. `samples = None, use_cache=True, analytic=False`
2. `samples = arr, use_cache=True, analytic=False`

In option 1, we don't input user generated samples. Instead, when the qnode is first evaluated on the probabilities, we do sampling (according to the `shots` argument which defaults to 1000). These samples should then be cached and re-used on subsequent evaluations. On the subsequent evaluations, we don't want to call `super().reset()` which gets rid of the samples, so we enter this if-statement and simply return nothing.

For 2, we straight away have user input samples that are added to `self.samples`, and we can enter this loop because again we don't want to run `super().reset()` and lose our samples.

TLDR: we don't want to call `super().reset()` else we lose our cached samples.

I think what is confusing is saying "There is no engine to reset in this case", since actually I think that is not true in case 1 as you point out. There, the engine is initialized but we don't need to worry about resetting it because it won't be used again.

I've changed the comment to:
`# In this case, we do not reset because we want to keep our samples`

trbromley:
image
Coverage looks good.

pennylane_sf/gbs.py
antalszava: > TLDR: we don't want to call super().reset() else we lose our cached samples.

Right! Actually, it can also be, that the `GBSDevice` grabs the important parts of `super().reset()` and for example never resets `self.samples`.

> don't need to worry about resetting it because it won't be used again.

Is this for sure? What if the user goes for `dev.use_cache=False` and runs another circuit? Although quite an edge case, maybe it's worth disabling the setting of `use_cache` (e.g. creating a `_use_cache` attribute that is set in `__init__` and making `use_cache` a property that returns `_use_cach`). Technically  `dev._use_cach` could be set by the user, but it would not be supported (and the user would have to tweak the code to produce unwanted behaviour).

trbromley: I like it!

trbromley: Added: 161e5c3516fc50c461273b1fc12bb5fa7d1053a5

antalszava (approved):

PR #56, opened 2020-10-02
Fix NumPy indexing bug

josh146:
A recent change in PennyLane resulted in indexing into NumPy arrays returning array scalars, rather than Python scalars:

```pycon
>>> from pennylane import numpy as np
>>> np.array([1, 2])[0]
array(1)
```

This causes an issue in the Strawberry Fields plugin, since the Hafnian functions provided by The Walrus do not expect scalar array inputs.

To fix this, we make sure that operation parameters are 'unwrapped' by the SF plugin before being sent to SF, by calling `p.unwrap()`.

**Note:** this can be considered a 'hotfix'. Ideally, this unwrapping should be done in PennyLane. However, there are a couple of reasons we don't do this here:

* This issue only affects non-tape mode. In tape mode, PennyLane _already_ unwraps the operation parameters. Once tape-mode is supported by default, this issue will no longer be applicable.

* We cannot do the unwrapping in the parent `Device` class, as there are some devices that might be in autograd passthru mode, and do not want unwrapping to occur.

### DIFF
--- a/pennylane_sf/simulator.py
+++ b/pennylane_sf/simulator.py
@@ -107,6 +107,7 @@ class StrawberryFieldsSimulator(Device):
 
         # translate to consecutive wires used by device
         device_wires = self.map_wires(wires)
+        par = [p.unwrap() if hasattr(p, "unwrap") else p for p in par]
 
         op = self._operation_map[operation](*sf_par)
         op | [self.q[i] for i in device_wires.labels]  # pylint: disable=pointless-statement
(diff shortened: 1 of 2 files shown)


thisac (approved):
:100: 

Should this be removed whenever tape-mode is default? If so, is it worth adding a TODO here?

josh146:
> Should this be removed whenever tape-mode is default? If so, is it worth adding a TODO here?

Good idea, I've added this :)

antalszava (approved):
Looks good to me! :100:

PR #59, opened 2020-12-19
Add Fock device support for FockDiagonalObservable

antalszava:
**Context**

Adds support for the Fock device to `FockDiagonalObservable`. For background description, see https://github.com/PennyLaneAI/pennylane/issues/951 and https://github.com/PennyLaneAI/pennylane/pull/969.


antalszava:
[ch1061]

PR #60, opened 2021-01-21
Fix tests with tape mode

thisac:
The tests are failing when enabling tape mode. The main thing changed is that `circuit.jacobian` is removed in favour of `qml.jacobian`.

Some issues that still aren't solved when enabling tape mode (that pass without):
* SF seems to support state preparation operations applied _after_ other operations (e.g. a two-mode sqeezing gate on `wires=[0, 1]` followed by a `FockStateVector` on `wires=0`). This causes an error in tape mode.
* Calculate the derivative of the `TensorN` operation causes an `IndexError: tuple index out of range` in autograd.
* Calculating the derivative of `ParamGraphEmbed` doesn't seem to work either.

**Benefits:**
The PennyLane-SF tests can pass in tape mode.

**Possible Drawbacks:**
Since I'm using `qml.jacobian` (instead of e.g. `circuit.qtape.jacobian`) the differentiation method cannot be chosen; thus it is currently always the default one used.

**Related GitHub Issues:**
Potentially related to some breaking demos which are using PennyLane-SF.

### DIFF
--- a/pennylane_sf/gbs.py
+++ b/pennylane_sf/gbs.py
@@ -28,6 +28,7 @@ from thewalrus.quantum import find_scaling_adjacency_matrix as rescale
 from thewalrus.quantum import photon_number_mean_vector
 
 import pennylane as qml
+from pennylane.tape import QuantumTape
 from pennylane.operation import Probability
 from pennylane.wires import Wires
 
@@ -246,7 +247,12 @@ class StrawberryFieldsGBS(StrawberryFieldsSimulator):
     def jacobian(self, *args):
         """Calculates the Jacobian of the device.
 
+        Either takes a single QuantumTape or the operations, observables and
+        variable_deps individually.
+
         Args:
+            tape (.QuantumTape): the tape to get the operations, observables and variable_deps
+                from, instead of passing them individually
             operations (list[~pennylane.operation.Operation]): operations to be applied to the
                 device
             observables (list[~pennylane.operation.Operation]): observables to be measured
@@ -258,6 +264,15 @@ class StrawberryFieldsGBS(StrawberryFieldsSimulator):
             array[float]: Jacobian matrix of size (``len(probs)``, ``num_wires``)
         """
         self.reset()
+
+        not_a_tape = len(args) == 1 and not isinstance(args[0], QuantumTape)
+        wrong_number_of_args = len(args) != 1 and len(args) != 3
+
+        if not_a_tape or wrong_number_of_args:
+            raise TypeError(
+                "Arguments must either contain a single QuantumTape or the operations"
+                "observables and variable_deps."
+            )
 
         if len(args) == 1:
             tape = args[0]
--- a/pennylane_sf/gbs.py
+++ b/pennylane_sf/gbs.py
@@ -261,10 +261,13 @@ class StrawberryFieldsGBS(StrawberryFieldsSimulator):
 
         if len(args) == 1:
             tape = args[0]
-            operations, observables, variable_deps = tape.operations, tape.observables, tape.graph.variable_deps
+            operations, observables, variable_deps = (
+                tape.operations,
+                tape.observables,
+                tape.graph.variable_deps,
+            )
         else:
             operations, observables, variable_deps = args
-
 
         requested_wires = observables[0].wires
 
--- a/tests/test_fock.py
+++ b/tests/test_fock.py
@@ -52,15 +52,6 @@ two_modes_single_real_parameter_gates = [
     ("ControlledPhase", qml.ControlledPhase),
 ]
 
-def get_classical_jacobian(qnode):
-    def classical_preprocessing(*args, **kwargs):
-        """Returns the trainable gate parameters for
-        a given QNode input"""
-        qnode.construct(args, kwargs)
-        return qml.math.stack(qnode.qtape.get_parameters())
-
-    return qml.jacobian(classical_preprocessing)
-
 
 # compare to reference SF engine
 def SF_gate_reference(sf_op, cutoff_dim, wires, *args):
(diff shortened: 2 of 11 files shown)


tests/test_fock.py
```diff
@@ -611,17 +611,14 @@ def circuit(r, phi):
         expected = np.exp(2 * r) * np.sin(phi) ** 2 + np.exp(-2 * r) * np.cos(phi) ** 2
         assert np.allclose(var, expected, atol=tol, rtol=0)
 
-        # circuit jacobians
-        gradA = circuit.jacobian([r, phi], method="A")
-        gradF = circuit.jacobian([r, phi], method="F")
+        grad = qml.jacobian(circuit)(r, phi)
```
josh146, line 614: Two mutually exclusive suggestions, not sure which one is the best one:

* you could make the replacement:

  ```python
  gradA = circuit.qtape.jacobian([r, phi], method="analytic")
  gradF = circuit.qtape.jacobian([r, phi], method="numeric")
  ```

* Alternatively, you can create two QNodes:

  ```python
  qnodeA = qml.QNode(circuit, dev, diff_method="parameter-shift")
  qnodeF = qml.QNode(circuit, dev, diff_method="finite-diff")
  gradA = qml.jacobian(qnodeA)(r, phi)
  gradF = qml.jacobian(qnodeF)(r, phi)
  ```

I'm not sure which is preferable 🤔 Maybe the first? I'm just hesitant to remove the finite difference check.

josh146 (commented):
Thanks @thisac for looking into this.

> Calculating the derivative of ParamGraphEmbed doesn't seem to work either.

Perhaps @trbromley might be able to quickly look at this one?

> SF seems to support state preparation operations applied after other operations (e.g. a two-mode sqeezing gate on wires=[0, 1] followed by a FockStateVector on wires=0). This causes an error in tape mode.

😬

tests/test_fock.py
thisac: Thanks @josh146! I briefly thought about the first option as well, but decided that it worked well enough just using `qml.jacobian`. But you're probably correct. Still WIP though, so I'll look into it more tomorrow. 🙂

pennylane_sf/gbs.py
```diff
@@ -259,6 +259,16 @@ def jacobian(self, operations, observables, variable_deps):
         """
         self.reset()
 
+        if len(args) == 1:
```
antalszava, line 262: Maybe worth using `qml.tape_mode_active()` here?

tests/test_gaussian.py
```diff
@@ -601,16 +601,14 @@ def circuit(r, phi):
         assert np.allclose(var, expected, atol=tol, rtol=0)
 
         # circuit jacobians
-        gradA = circuit.jacobian([r, phi], method="A")
-        gradF = circuit.jacobian([r, phi], method="F")
+        grad = qml.jacobian(circuit)(r, phi)
```
antalszava, line 604: It might be worth creating the qnodes explicitly such that both the analytic and finite diff checks can be kept

antalszava (commented):
Had a skim through, looks like it's coming along! 🚀

tests/test_gaussian.py
thisac: Yes, the plan is to go back to calculating both analytic and finite diffs (that's why I noticed the difference between `qml.jacobian` and `tape.jacobian`. :slightly_smiling_face:

pennylane_sf/gbs.py
thisac: I agree it might seem a bit loose to only check for the number of args, but it was the best way I could think of solving this while still keeping compatibility.

antalszava: How about raising an error if
1) `len(args) == 1` and `args[0]` is not a tape and
2) `len(args) != 1` or `len(args) != 3`?

thisac: Sure, I could add something like that. Makes sense! :+1:

thisac:
Thanks @josh146 and @antalszava  for the comments. I've updated the tests, and everything now passes when running against PennyLane master. :tada: I guess we better wait to merge this until after the PL release (since the tests won't pass until then).

pl-sf-tests

tests/test_fock.py
```diff
@@ -745,13 +753,21 @@ def circuit(a, phi):
...
+
         # differentiate with respect to parameter a
-        res_F = circuit.jacobian([a, phi], wrt={0}, method="F").flat
+        circuit.qtape.trainable_params = {0}
+        res_F = circuit.qtape.jacobian(dev, method="numeric").flatten()
         expected_gradient = 2 * np.exp(-(a ** 2)) * a ** (2 * n - 1) * (n - a ** 2) / fac(n)
         assert np.allclose(res_F, expected_gradient, atol=tol, rtol=0)
 
+        # re-construct tape to reset trainable_params
+        circuit.construct([a, phi], {})
```
thisac, line 766: I was a bit unsure whether this was the best way to do this, or whether this is actually something that should be fixed in PL core. The issue being that when setting e.g. `circuit.qtape.trainable_params = {0}` above and then calling `circuit.qtape.jacobian`, the second parameter (`1`) is given grad method `None`. Then, attempting to set `circuit.qtape.trainable_params = {1}` later doesn't work, since it only sets the grad method if the parameter has none from before. 

This happens in `JacobianTape._grad_method_validation`:

```python
if "grad_method" not in self._par_info[0]:
        self._update_gradient_info()
```

Reconstructing the tape was the easiest way of fixing this (other that perhaps removing the `if "grad_method" not in self._par_info[0]:` above).

josh146 (commented):
@thisac I noticed that the TF tests are not passing, so made a quick PR into this branch to fix it: #62

> I guess we better wait to merge this until after the PL release (since the tests won't pass until then).

I've updated the `requirements.txt` to temporarily install PL from github. Let's get this merged ASAP, since it is blocking other work (e.g., the PL-SF demos).

josh146 (approved):

PR #65, opened 2021-04-12
Remove analytic keyword argument and v0.15 version bump

josh146:


.github/CHANGELOG.md
```diff
@@ -1,19 +1,18 @@
...
 
-### Improvements
-
-### Documentation
-
-### Bug fixes
+* For compatibility with PennyLane v0.15, the `analytic` keyword argument
+  has been removed from all devices. Analytic expectation values can
+  still be computed by setting `shots=None`.
+  [(#62)](https://github.com/XanaduAI/pennylane-sf/pull/62)
```
josh146, line 8:
```suggestion
  [(#65)](https://github.com/XanaduAI/pennylane-sf/pull/65)
```

pennylane_sf/fock.py
```diff
@@ -124,8 +122,8 @@ class StrawberryFieldsFock(StrawberryFieldsSimulator):
 
     _circuits = {}
 
-    def __init__(self, wires, *, analytic=True, cutoff_dim, shots=1000, hbar=2):
-        super().__init__(wires, analytic=analytic, shots=shots, hbar=hbar)
+    def __init__(self, wires, *, shots=None, cutoff_dim, hbar=2):
```
trbromley, line 125: :thinking: Is it allowed to have a non-default argument (`cutoff_dim`) after a default one?

pennylane_sf/gbs.py
```diff
@@ -24,11 +24,10 @@
 import strawberryfields as sf
 from strawberryfields.ops import GraphEmbed, MeasureFock
 from strawberryfields.utils import all_fock_probs_pnr
-from thewalrus.quantum import find_scaling_adjacency_matrix as rescale
+from thewalrus.quantum import adj_scaling as rescale
```
trbromley, line 27: Is this due to an update to `thewalrus`?

pennylane_sf/fock.py
trbromley: (from some local testing): ohhh, so the `*` is key? In doing so, we are saying that `wires` is the only positional argument, and while `shots` and `hbar` have defaults, `cutoff_dim` is a non-default keyword argument?

pennylane_sf/gbs.py
```diff
@@ -76,10 +73,8 @@ class StrawberryFieldsGBS(StrawberryFieldsSimulator):
 
     _capabilities = {"model": "cv", "provides_jacobian": True}
 
-    def __init__(
-        self, wires, *, analytic=True, cutoff_dim, shots=1000, use_cache=False, samples=None
-    ):
-        super().__init__(wires, analytic=analytic, shots=shots)
+    def __init__(self, wires, *, cutoff_dim, shots=None, use_cache=False, samples=None):
```
trbromley, line 76: In the others we changed the order to `shots, cutoff_dim`, do we want to do the same here?

pennylane_sf/gbs.py
```diff
@@ -244,53 +239,23 @@ def probability(self, wires=None):
...
-            tape = args[0]
-            operations, observables, variable_deps = (
-                tape.operations,
-                tape.observables,
-                tape.graph.variable_deps,
-            )
-        else:
-            operations, observables, variable_deps = args
-
+        operations, observables = (tape.operations, tape.observables)
```
trbromley, line 252: Nice to make this smaller!

tests/test_gbs.py
```diff
@@ -374,25 +373,25 @@ class TestStrawberryFieldsGBS:
...
         assert dev.cutoff == cutoff
-        assert dev.analytic is analytic
+        assert dev.analytic is (shots is None)
         assert dev.use_cache is use_cache
 
     def test_load_device_with_samples(self):
         """Test that the device loads correctly with input samples"""
         a = np.ones((10, 2))
-        dev = qml.device("strawberryfields.gbs", wires=2, cutoff_dim=3, analytic=False, samples=a)
+        dev = qml.device("strawberryfields.gbs", wires=2, cutoff_dim=3, shots=10, samples=a)
```
trbromley, line 394: This might highlight a subtlety in the device that `shots` doesn't need to be set to 10. It could be set to anything as long as `shots is not None` (i.e., we are in "non-analytic mode").

Overall, I think this device did lean on the `analytic` keyword argument a bit more than usual. It is probably ok though - not sure if this device is used much and things still work if you set shots to be the number of samples (as you have done above)

tests/test_gbs.py
```diff
@@ -581,7 +580,7 @@ def test_probability_non_analytic_all_wires(self):
         """Test that the probability method in non-analytic mode returns the expected dictionary
         mapping from samples to probabilities when a fixed set of pre-generated samples are
         input. The lexicographic order of the keys in the returned dictionary is also checked."""
-        dev = qml.device("strawberryfields.gbs", wires=4, cutoff_dim=3, analytic=False)
+        dev = qml.device("strawberryfields.gbs", wires=4, cutoff_dim=3, shots=4)
```
trbromley, line 583: E.g. here, `shots=4` but `len(samples) = 10`. It probably works fine because `shots` is just ignored aside from establishing that we're in non-analytic mode :thinking:

trbromley (changes requested):
Thanks @josh146! Looks great, one requested change would be to update the mentions of `analytic` in the docs:
- https://pennylane-sf.readthedocs.io/en/latest/devices/gbs.html
- https://pennylane-sf.readthedocs.io/en/latest/devices/tf.html

pennylane_sf/fock.py
josh146: Yep the `*` is key!

I don't remember the reason why the SF plugin is like this 😆

pennylane_sf/gbs.py
josh146: yep, I was getting a deprecation notice 😆

tests/test_gbs.py
josh146: Yeah, I noticed this while updating the tests. Did you want to add validation to ensure that `shots==len(samples)`, or do you want to leave it as is?

Alternatively, you could allow `shots<=len(samples)`, and have it truncate the samples if shorter.

tests/test_gbs.py
josh146: Yep - I wasn't sure whether this was a 'bug' or not, since this was the same behaviour _before_ the shots refactor - hence why I left it as is.

pennylane_sf/gbs.py
josh146: Sure! I'll do it now.

josh146:
Thanks @trbromley, good catch with the docs! Have updated all device doc pages.

tests/test_fock.py
```diff
@@ -88,7 +88,7 @@ def test_load_fock_device(self):
         assert dev.num_wires == 2
         assert dev.cutoff == 5
         assert dev.hbar == 2
-        assert dev.shots == 1000
+        assert dev.shots == None
```
trbromley, line 91:
```suggestion
        assert dev.shots is None
```

tests/test_gaussian.py
```diff
@@ -93,7 +93,7 @@ def test_load_gaussian_device(self):
         dev = qml.device("strawberryfields.gaussian", wires=2, cutoff_dim=15)
         assert dev.num_wires == 2
         assert dev.hbar == 2
-        assert dev.shots == 1000
+        assert dev.shots == None
```
trbromley, line 96:
```suggestion
        assert dev.shots is None
```

tests/test_tf.py
```diff
@@ -90,7 +90,7 @@ def test_load_tf_device(self):
         assert dev.num_wires == 2
         assert dev.cutoff == 5
         assert dev.hbar == 2
-        assert dev.shots == 1000
+        assert dev.shots == None
```
trbromley, line 93:
```suggestion
        assert dev.shots is None

```

trbromley (approved):
Thanks @josh146!

PR #71, opened 2021-06-21
Adds the compatibility tag and runs the tests on Python 3.9

rmoyard:

- Add support and test for Python 3.9.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -3,7 +3,7 @@
 ### Breaking changes
 
 * Add support and tests for Python 3.9.
-  [(#71)](https://github.com/PennyLaneAI/pennylane-sf/pull/70)
+  [(#71)](https://github.com/PennyLaneAI/pennylane-sf/pull/71)
   
 ### Bug fixes
 
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Release 0.16.0
 
-### Breaking changes
+### Improvements
 
 * Adds the compatibility tag and runs the tests on Python 3.9.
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -1,3 +1,3 @@
 strawberryfields>=0.15
-pennylane>=0.15
+pennylane>=0.16
```
antalszava, line 2: @rmoyard we don't need to pin here (that was only done in the Cirq plugin).
```suggestion
pennylane>=0.15
```

setup.py
```diff
@@ -23,7 +23,7 @@
 
 requirements = [
     "strawberryfields>=0.15",
-    "pennylane>=0.15"
+    "pennylane>=0.16"
```
antalszava, line 26:
```suggestion
    "pennylane>=0.15"
```

.github/CHANGELOG.md
```diff
@@ -1,4 +1,10 @@
 # Release 0.16.0
+
+### Breaking changes
+
+* Add support and tests for Python 3.9.
```
antalszava, line 5: Also, would suggest moving this to `Improvements`.
```suggestion
* Adds the compatibility tag and runs the tests on Python 3.9.
```

antalszava (approved):
:100: Thank you!

.github/CHANGELOG.md
```diff
@@ -1,4 +1,10 @@
 # Release 0.16.0
+
+### Improvements
+
+* Adds the compatibility tag and runs the tests on Python 3.9.
```
antalszava, line 5:
```suggestion
* Adds the compatibility tag for Python 3.9.
```

PR #77, opened 2021-10-08
Interferometer deprecation in PennyLane

rmoyard:

The recent deprecation of `qml.Interferometer` to `qml.InterferometerUnitary` implies changes in the mapping of operation with StrawberryFields.

**Related GitHub Issues:**

https://github.com/PennyLaneAI/pennylane/issues/1256

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -6,12 +6,12 @@
 
 ### Improvements
 
+* Change the name of `qml.Interferometer` to `qml.InterferometerUnitary` due to PennyLane deprecation
+  [(#77)](https://github.com/PennyLaneAI/pennylane-sf/pull/77)
+
 ### Documentation
 
 ### Bug fixes
-
-* Change the name of `qml.Interferometer` to `qml.InterferometerUnitary` due to PennyLane deprecation
-  [(#77)](https://github.com/PennyLaneAI/pennylane-sf/pull/77)
 
 ### Contributors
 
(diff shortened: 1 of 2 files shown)


pennylane_sf/tf.py
```diff
@@ -190,7 +190,7 @@ class StrawberryFieldsTF(StrawberryFieldsSimulator):
     def __init__(self, wires, *, cutoff_dim, shots=None, hbar=2):
         super().__init__(wires, shots=shots, hbar=hbar)
         self.cutoff = cutoff_dim
-        self.params = dict()
+        self.params = {}
```
antalszava, line 193: How come these were changed? Codefactor?

.github/CHANGELOG.md
```diff
@@ -10,10 +10,15 @@
 
 ### Bug fixes
 
+* Change the name of `qml.Interferometer` to `qml.InterferometerUnitary` due to PennyLane deprecation
```
antalszava, line 13: Maybe this is better placed in `Improvements`? :thinking: It doesn't feel like a bug fix (as it's a change due to intended behaviour), but also not like any of the other categories :smile:

.github/CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-# Release 0.17.0-dev
+# Release 0.19.0-dev
```
antalszava, line 1: Nice! :) Likely the version in `_version.py` also needs to be updated.

pennylane_sf/tf.py
rmoyard: Exactly haha

antalszava (approved):
Looks good to me! :slightly_smiling_face: Thanks for getting this through so quickly. :rocket:

PR #80, opened 2021-11-01
Release pl first assump in vbscript

Jaybsoni:
Updated the automated version bump code to support version bumps both before and after a pennylane core release

### DIFF
--- a/.github/workflows/vb_script.py
+++ b/.github/workflows/vb_script.py
@@ -16,7 +16,7 @@ import pennylane as qml
 pl_version = '"' + qml.version() + '"'
 
 
-def bump_version(version_line, pre_release, pl_release):
+def bump_version(version_line, pre_release):
     """ A helper function which takes the current version string and
     replaces it with the bumped version depending on the pre/post
     release flag.
@@ -26,8 +26,6 @@ def bump_version(version_line, pre_release, pl_release):
             version of the plugin.
          pre_release (bool): A flag which determines if this is a
             pre-release or post-release version bump.
-        pl_release (bool): A flag which determines if the new
-            version of PennyLane has already been released to Pypi
 
     Returns:
         resultant_line (string): A string of the same form as the version line
@@ -40,10 +38,6 @@ def bump_version(version_line, pre_release, pl_release):
     if pre_release:
         curr_version = pl_version  # get current Pennylane version
 
-        if pl_release:
-            data[-1] = curr_version
-            return " ".join(data), curr_version  # already bumped
-
     split_version = curr_version.split(".")  # "0.17.0" --> ["0,17,0"]
     split_version[1] = str(int(split_version[1]) + 1)  # take middle value and cast as int and bump it by 1
 
@@ -55,15 +49,13 @@ def bump_version(version_line, pre_release, pl_release):
     return " ".join(data), bumped_version
 
 
-def update_version_file(path, pre_release=True, pl_release_status=False):
+def update_version_file(path, pre_release=True):
     """ Updates the __version__ attribute in a specific version file.
 
     Args:
         path (str): The path to the version file.
         pre_release (bool): A flag which determines if this is a
             pre-release or post-release version bump.
-        pl_release_status (bool): A flag which determines if the new
-            version of PennyLane has already been released to Pypi
 
     Return:
         new_version (str): The bumped version string.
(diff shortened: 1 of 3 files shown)


.github/workflows/vb_script.py
```diff
@@ -159,7 +169,11 @@ def update_changelog(path, new_version, pre_release=True):
                         help="True if this is a pre-release version bump, False if it is post release")
     parser.add_argument("--post_release", dest="release_status", action="store_false",
                         help="True if this is a pre-release version bump, False if it is post release")
+    parser.add_argument("--post_pl_release", dest="pl_release_status",
+                        action="store_true", help="True if PL has already been released")
+    parser.add_argument("--pre_pl_release", dest="pl_release_status",
+                        action="store_false", help="False if PL has not been released")
```
Jaybsoni, line 175: Added new parameter to track if this version bump is happening before or after a core PL release, and working accordingly

antalszava: As mentioned in the Lightning version bump PR, not sure if we'll need this option.

For now, we have the following release order:

0. Pre-release version bump for PennyLane and all plugins (including PennyLane-Lightning)
1. PennyLane-Lightning release
2. PennyLane release
3. Every other plugin release

As we'd have to change the code if this order changes anyway, maybe it's okay to just leave a comment at the part of the code where it's relevant.

.github/workflows/vb_script.py
```diff
@@ -36,13 +36,15 @@ def bump_version(version_line, pre_release):
     curr_version = data[-1]
 
     if pre_release:
-        bumped_version = pl_version  # get current Pennylane version
-    else:
-        split_version = curr_version.split(".")  # "0.17.0" --> ["0,17,0"]
-        split_version[1] = str(int(split_version[1]) + 1)  # take middle value and cast as int and bump it by 1
+        curr_version = pl_version  # get current Pennylane version
```
antalszava, line 39: Just to make sure: this code was changed because previously it assumed that PennyLane would be released before we make the pre-release version bump of plugins?

.github/workflows/vb_script.py
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 import argparse
 import pennylane as qml
-pl_version = '"' + qml.version() + '"\n'  # we expect PL v0.X.0 here if we're releasing v0.X.0 of the plugin 
+pl_version = '"' + qml.version() + '"'
```
antalszava, line 16: Maybe worth leaving a modified comment in?

`# we expect PL v0.X.0 here if we're releasing v0.X+1.0 of the plugin`

.github/workflows/vb_script.py
```diff
@@ -36,13 +36,15 @@ def bump_version(version_line, pre_release):
     curr_version = data[-1]
```
antalszava, line 36: Perhaps it can be helpful to include an example in the docstring for what this function does in the pre-release and post-release cases.
```
Assume that we have PennyLane version 0.18.0 as the latest release on PyPi.

Pre-release:

We gather the version number for the latest of PennyLane release (i.e., v0.18.0) and increment the middle value to obtain the version of the plugin that we'll release: v0.19.0.

Post-release:

We have just released the v0.19.0 version of the plugin and have v0.19.0 set as the version number. We are entering a new development phase of the plugin attached to the next release, so we increment the version number (0.20.0) and add the "-dev" suffix (0.20.0-dev).
```

antalszava (commented):
Just a couple of comments. :+1:

.github/workflows/vb_script.py
Jaybsoni: Yes the logic is such that it can handle the following cases: 
1. 
    - Pre-release version bump plugin
    - Release PL 
    - Release Plugin
    - Post-release version bump plugin 

2.
    - Pre-release version bump plugin 
    - Release plugin 
    - Post-release version bump plugin 
    - Release PL

I think there was a mis-communication before because this functionality was already present before. For some reason I thought you wanted to also include the following case: 

- Release PL 
- Pre-release version bump plugin 
- Release plugin 
- Post-release version bump plugin

.github/workflows/vb_script.py
Jaybsoni:
```suggestion
pl_version = '"' + qml.version() + '"' # we expect PL v0.X.0 here if we're releasing v0.X+1.0 of the plugin
```

.github/workflows/vb_script.py
antalszava: Oh, sorry about that! Sure, 1. sounds good.

Previously, the only other variant we've seen examples for was (e.g., now for Lightning):

* Pre-release version bump plugin
* Release PL
* Release Plugin
* Post-release version bump plugin

.github/workflows/vb_script.py
Jaybsoni: Got it

antalszava (approved):
Looks good to me! :muscle:

antalszava:
[sc-10712]

PR #84, opened 2021-11-11
Use `qml.probs` instead of `qml.Identity` for the gbs device

antalszava:


pennylane_sf/gbs.py
```diff
@@ -253,8 +252,7 @@ def jacobian(self, tape):
         requested_wires = observables[0].wires
 
         # Create dummy observable to measure probabilities over all wires
-        obs_all_wires = qml.Identity(wires=self.wires)
-        obs_all_wires.return_type = Probability
+        obs_all_wires = qml.probs(wires=self.wires)
```
Jaybsoni, line 255: Looks much cleaner now 👍🏼

Jaybsoni (approved):

PR #85, opened 2021-11-11
Probs tests

antalszava:


requirements.txt
```diff
@@ -1,3 +1,3 @@
 strawberryfields>=0.15
-pennylane>=0.19
+git+git://github.com/PennyLaneAI/pennylane
```
antalszava, line 2: This change will be reverted right before merging. The idea was to check that tests pass with the latest version of PennyLane. Certain was were failing ([see here](https://github.com/PennyLaneAI/plugin-test-matrix/runs/4182515243?check_suite_focus=true#step:8:46)).
```suggestion
pennylane>=0.19
```

Jaybsoni: Ahh makes sense

tests/test_tf.py
```diff
@@ -513,7 +513,7 @@ def test_trace(self, tol):
         def circuit(x, y):
             qml.Squeezing(x, 0, wires=0)
             qml.Squeezing(y, 0, wires=1)
-            return qml.expval(qml.Identity(wires=[0, 1]))
+            return qml.expval(qml.Identity(wires=[0])), qml.expval(qml.Identity(wires=[1]))
```
Jaybsoni, line 516: Yup ! because it only takes a single wire now

Jaybsoni (approved):

antalszava:
With #86 merged, closing this.

PR #86, opened 2021-12-07
Fix tests

puzzleshark:
# plugin is failing due to api changes in strawberry fields

### DIFF
--- a/pennylane_sf/expectations.py
+++ b/pennylane_sf/expectations.py
@@ -131,7 +131,7 @@ def fock_state(state, device_wires, params):
     if isinstance(state, BaseFockState):
         # fock state
         dm = state.reduced_dm(modes=device_wires.tolist())
-        ex = dm[tuple([n[i // 2] for i in range(len(n) * 2)])].real
+        ex = dm[tuple(n[i // 2] for i in range(len(n) * 2))].real
 
     elif isinstance(state, BaseGaussianState):
         # Reduced Gaussian state
--- a/pennylane_sf/expectations.py
+++ b/pennylane_sf/expectations.py
@@ -62,11 +62,6 @@ def identity(state, device_wires, params):
     N = state.num_modes
     D = state.cutoff_dim
 
-    if N == len(device_wires):
-        # trace of the entire system
-        tr = state.trace()
-        return tr, tr - tr ** 2
-
     # get the reduced density matrix
     N = len(device_wires)
     dm = state.reduced_dm(modes=device_wires.tolist())
--- a/pennylane_sf/remote.py
+++ b/pennylane_sf/remote.py
@@ -111,14 +111,14 @@ class StrawberryFieldsRemote(StrawberryFieldsSimulator):
 
         if isinstance(wires, int):
             raise ValueError(
-                "Device has a fixed number of {} modes. The wires argument can only be used "
-                "to specify an iterable of wire labels.".format(self.num_wires)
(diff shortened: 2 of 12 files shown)


pennylane_sf/remote.py
```diff
@@ -24,6 +24,7 @@
 import strawberryfields as sf
 
 # import state preparations, gates and measurements
+import xcc
```
antalszava, line 27: Will we need to add `xcc` as a dependency? Or it's okay as it ships with Strawberry Fields?

josh146: Best to let it come via SF (since SF might want to pin XCC to a particular version).

However, there is a subtlety --- only the latest version of SF depends on XCC, so a user with an old SF version pre-installed might have an issue when installing this plugin.

A solution could be to increase the pinning of SF in the plugin `setup.py`, to the latest one?

antalszava:
Hey Sam, the remaining issue with CodeCov will be due to the case of identity computation that is not covered anymore by the tests.

pennylane_sf/remote.py
puzzleshark: pinned SF to the latest in `setup.py` as suggested

pennylane_sf/expectations.py
```diff
@@ -62,11 +62,6 @@ def identity(state, device_wires, params):
     N = state.num_modes
     D = state.cutoff_dim
 
-    if N == len(device_wires):
-        # trace of the entire system
-        tr = state.trace()
-        return tr, tr - tr ** 2
-
```
antalszava, line 69: Just a note here: this is being removed because:
1. Only a single wire can be specified for `qml.Identity` with the latest changes to PennyLane;
2. The SF devices do not support tensor products.

pennylane_sf/remote.py
```diff
@@ -110,21 +111,21 @@ def __init__(self, *, backend, wires=None, cutoff_dim=5, shots=1000, hbar=2, sf_
...
+                f"Device has a fixed number of {self.num_wires} modes and "
+                f"cannot be created with {len(wires)} wires."
             )
 
         super().__init__(wires, shots=shots, hbar=hbar)
         self.eng = eng
 
         if sf_token is not None:
-            sf.store_account(sf_token)
+            xcc.Settings(REFRESH_TOKEN=sf_token).save()
```
antalszava, line 128: Nice :+1:

setup.py
```diff
@@ -22,7 +22,7 @@
 
 
 requirements = [
-    "strawberryfields>=0.15",
+    "strawberryfields>=0.19",
```
antalszava, line 25: Should we specify the minor version here too?
```suggestion
    "strawberryfields>=0.19.2",
```

tests/test_fock.py
```diff
@@ -598,7 +598,7 @@ def test_first_order_cv(self, tol):
         """Test variance of a first order CV expectation value"""
         dev = qml.device("strawberryfields.fock", wires=1, cutoff_dim=15)
 
-        @qml.qnode(dev)
+        @qml.qnode_old.qnode(dev)
```
antalszava, line 601: Note: the changes to the new QNode do not seem to be compatible with the GBS device, hence we resort to using the old QNode.

antalszava (approved):
Looks good to me :muscle: Nice set of fixing :slightly_smiling_face:

PR #93, opened 2022-02-01
fixed tests requiring gradients

puzzleshark:
Do to recent [refactor](https://github.com/PennyLaneAI/pennylane/commit/bc822b40d97c11fc59570f96d4280374cee639ac) of PennyLane, default behaviour of `requires_grad` some scalar values need to be explicitly set as autograd tensors, with `requires_grad=True`.

### DIFF
--- a/pennylane_sf/expectations.py
+++ b/pennylane_sf/expectations.py
@@ -68,9 +68,9 @@ def identity(state, device_wires, params):
 
     # construct the standard 2D density matrix, and take the trace
     new_ax = np.arange(2 * N).reshape([N, 2]).T.flatten()
-    tr = np.trace(dm.transpose(new_ax).reshape([D ** N, D ** N])).real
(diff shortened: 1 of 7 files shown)


antalszava (approved):
Looks good! :tada: Thank you!

For some reason, we don't seem to be running the device tests that we run in the plugin test matrix. :thinking: Do the tests pass now well locally with these changes?

antalszava:
@Jaybsoni it looks like the only changes affecting the logic were made to the test files. Therefore, we would not need to release this plugin.

Jaybsoni:
> @Jaybsoni it looks like the only changes affecting the logic were made to the test files. Therefore, we would not need to release this plugin.

Sounds good ! 
So we only need to version bump the Qiskit plugin

antalszava:
[sc-14395]

PR #94, opened 2022-02-02
fix requirements

puzzleshark:
# revert dependencies back to original


antalszava (approved):
@puzzleshark thank you for catching this! :slightly_smiling_face:

PR #103, opened 2022-03-28
Add Python 3.10 support

antalszava:


thisac (approved):
Looks good. Noticed that Python 3.6 is mentioned as supported in the readme still. Could perhaps be removed here as well (along with bumping the SF and PL versions there as well).

antalszava:
Oh great catch, thank you!

PR #110, opened 2022-04-21
Added deprecation warning for gbs device

puzzleshark:


pennylane_sf/gbs.py
```diff
@@ -82,6 +85,9 @@ def __init__(self, wires, *, shots=None, cutoff_dim, use_cache=False, samples=No
         self._WAW = None
         self.Z_inv = None
 
+        warnings.warn("strawberryfields.gbs will be deprecated for pennylane>=v0.23.0."
```
antalszava, line 88:
```suggestion
        warnings.warn("strawberryfields.gbs is being deprecated and won't work as expected with PennyLane>=v0.23.0."
```

pennylane_sf/gbs.py
```diff
@@ -82,6 +85,9 @@ def __init__(self, wires, *, shots=None, cutoff_dim, use_cache=False, samples=No
...
+                      "If you would like to use pennylane must be <=v022.0", DeprecationWarning)
```
antalszava, line 89: Maybe we can also encourage users to reach out if they want to keep using this device.
```suggestion
                      "If you would like to use this device now and in the future, downgrade PennyLane to <=v022.0 and create a related issue on the GitHub page of PennyLane-SF.", DeprecationWarning)
```

antalszava:
@puzzleshark this is looking nice! :) How come the coverage doesn't drop with the tests being removed? Lmk if I should go for a review.

puzzleshark:
> @puzzleshark this is looking nice! :) How come the coverage doesn't drop with the tests being removed? Lmk if I should go for a review.

ya a review would be great! Not sure why the coverage is not dropping. It's a good question

PR #125, opened 2022-12-14
Convert list index to a tuple

antalszava:
With a newer version of NumPy, a list containing indices of a NumPy array may result in errors if the list was created using `slice` objects.

This PR converts the list to a tuple, which seems to work great even with newer NumPy versions.


Jaybsoni (commented):
Elegant 💯

Jaybsoni (approved):

PR #129, opened 2023-02-28
remove python 3.7 support

timmysilv:
Locally, I did the following:

```bash
# test with minimum versions in setup.py
pip install pennylane==0.19.0 strawberryfields==0.22.0 tensorflow-macos pytest pytest-mock
pip install -r requirements.txt
pip install -e .
make test

# test with latest versions of unpinned libs (from requirements-ci.txt)
pip install pennylane==0.29.0 strawberryfields==0.23.0
make test
```

Both passed fine. Won't hard-pin anything to recent versions because they're all compatible! I can bump the pinned ones in `requirements.txt` if anyone wants.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Breaking changes
 
 * Remove python 3.7 support.
+  [(#129)](https://github.com/PennyLaneAI/pennylane-sf/pull/129)
 
 ### Improvements
 
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Breaking changes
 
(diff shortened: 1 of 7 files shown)


timmysilv:
I think I'll need admin support because of the `tests (3.7)` CI check that needs an override while it's still in master?

.github/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### Breaking changes
 
+* Remove python 3.7 support.
```
timmysilv, line 7:
```suggestion
* Remove python 3.7 support.
  [(#129)](https://github.com/PennyLaneAI/pennylane-sf/pull/129)
```

albi3ro (approved):

PR #147, opened 2023-06-26
Switch to GA4

doctorperceptron:
Google is deprecating Universal Analytics on July 1st. This PR switches to the newer GA4 tag.


timmysilv (approved):
🎉

PR #151, opened 2023-12-18
Update tutorial image links

ottoak:
Image links pointed to old locations in the QML repo and one was broken:
image

This PR fixes these links to point to the right locations

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -5,7 +5,7 @@ sphinx:
 
 python:
     install:
-        - requirements: ci_build_requirements.txt
+        - requirements: requirements-ci.txt
         - requirements: doc/requirements.txt
         - method: pip
           path: .
@@ -17,4 +17,4 @@ build:
         python: "3.8"
     jobs:
         pre_install:
-            - echo "setuptools~=66.0\npip~=22.0" >> ci_build_requirements.txt
+            - echo "setuptools~=66.0\npip~=22.0" >> requirements-ci.txt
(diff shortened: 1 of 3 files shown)


josh146:
thanks @ottoak! 

@lillian542 looks like this package will need to be pinned to an earlier version of TF, since TF 2.13 is incompatible with PL 0.29 (which this package pins to).

So we may need a new release of this plugin, with TF pinned

albi3ro:
> thanks @ottoak!
> 
> @lillian542 looks like this package will need to be pinned to an earlier version of TF, since TF 2.13 is incompatible with PL 0.29 (which this package pins to).
> 
> So we may need a new release of this plugin, with TF pinned

Python doesn't really have a system for pinning optional dependencies, and I don't think we want to make TF a hard dependency.  If the concern right now is the CI, we can pin the version in `requirements-ci.txt`.

This long-standing problem of not knowing which combination of versions worked at a particular point in time will potentially be fixed by adding poetry support in the new year, but until then... I think the best we can do is update the requirements files.

albi3ro:
Any idea how to fix this issue?

Screenshot 2023-12-19 at 9 35 07 AM

josh146:
@albi3ro oops good point, I forgot TF was an optional dependency. In that case, should be fine to just pin it in thebCI requirements.

Regarding RTDs, could we copy a RTDs config file from a more up to date plugin?

lillian542 (approved):

ottoak:
@josh146 @lillian542 @albi3ro This looks good to go now, but I don't have merge privileges - can one of you do so?

